PR #3, opened 2018-03-21
Model: add predicate

irina-dragoste:
review request for adding Predicate(String name, int arity) model


vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/impl/Expressions.java
```diff
@@ -54,35 +55,84 @@ public static Variable makeVariable(String name) {
...
-	 *            non-blank predicate name
+	 *            a non-null {@link Predicate}
 	 * @param terms
-	 *            non-empty array of non-null terms
+	 *            non-empty, non-null array of non-null terms. Aray size must be the
+	 *            same as the given {@code predicate} arity.
 	 * @return an {@link Atom} corresponding to the input
 	 */
-	public static Atom makeAtom(final String predicate, final Term... terms) {
+	public static Atom makeAtom(final Predicate predicate, final Term... terms) {
```
mkroetzsch: I would suggest to keep this method using a predicate name string. It's a convenience method that should be as simple as possible.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/impl/PredicateImpl.java
```diff
@@ -0,0 +1,74 @@
...
+		if (this == obj) {
+			return true;
+		}
+		if (obj == null) {
+			return false;
+		}
+		if (!(obj instanceof PredicateImpl)) {
+			return false;
+		}
+		final PredicateImpl other = (PredicateImpl) obj;
```
mkroetzsch: The above should be changed so that objects of two Predicate implementations can still be equal (check and cast to Predicate, not to PredicateImpl)

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/ReasonerImpl.java
```diff
@@ -139,7 +140,7 @@ public void load() throws AlreadyStartedException, EDBConfigurationException, IO
 			edbPredicates.add(edbPredConfig.getPredicate());
 		}
 		for (final Atom fact : this.facts) {
-			edbPredicates.add(fact.getPredicate());
+			edbPredicates.add(fact.getPredicate().getName());
```
mkroetzsch: Here and below in this file: Shouldn't we just use Predicate objects here instead of strings?

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/ReasonerImpl.java
```diff
@@ -169,21 +170,21 @@ private String edbPredicatesConfigToString() {
...
 		for (final Atom fact : this.facts) {
 			factsMap.putIfAbsent(fact.getPredicate(), new ArrayList<>());
 			factsMap.get(fact.getPredicate()).add(fact.getTerms());
 		}
-		for (final String predName : factsMap.keySet()) {
-			final List<List<Term>> predNameArgs = factsMap.get(predName);
-			final int arity = predNameArgs.get(0).size();
-			final String[][] tuplesMatrix = new String[predNameArgs.size()][arity];
-			for (int i = 0; i < predNameArgs.size(); i++) {
+		for (final Predicate pred : factsMap.keySet()) {
```
mkroetzsch: style: call variables like their class, i.e., "predicate" instead of "pred"

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/model/AtomImplTest.java
```diff
@@ -29,78 +30,103 @@
...
-		Variable x = Expressions.makeVariable("X");
-		Constant c = Expressions.makeConstant("c");
-		Atom atom1 = Expressions.makeAtom("p", Arrays.asList(x, c));
-		Atom atom2 = Expressions.makeAtom("p", x, c);
-		Atom atom3 = new AtomImpl("q", Arrays.asList(x, c));
-		Atom atom4 = new AtomImpl("p", Arrays.asList(c, x));
+		final Variable x = Expressions.makeVariable("X");
+		final Constant c = Expressions.makeConstant("c");
+
+		final Predicate predicateP = new PredicateImpl("p", Arrays.asList(x, c).size());
```
mkroetzsch: "Arrays.asList(x, c).size()" should be replaced by 2! (also below)

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/model/AtomImplTest.java
```diff
@@ -29,78 +30,103 @@
...
 	@Test(expected = IllegalArgumentException.class)
-	public void temrsNonEmpty() {
+	public void termsNonEmpty() {
 		Expressions.makeAtom("p");
 	}
 	
 	@Test(expected = NullPointerException.class)
 	public void predicateNotNull() {
-		Expressions.makeAtom(null, Expressions.makeConstant("c"));
+		final Predicate nullPredicate = null;
```
mkroetzsch: what is the use of assigning null to a variable instead of just using it directly?

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/impl/Expressions.java
irina-dragoste: we do (it's the method above, with String predicate renamed to predicateName)

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/model/AtomImplTest.java
irina-dragoste: the call  makeAtom(null, Constant constant) is ambiguous, the first argument can be a String (predicate name) or a Predicate

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/ReasonerImpl.java
irina-dragoste: I will fix this in the "query" branch

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/ReasonerImpl.java
irina-dragoste: I will fix this in the "query" branch

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/impl/PredicateImpl.java
irina-dragoste: done in the next commit

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/model/AtomImplTest.java
irina-dragoste: done in the next commit

PR #45, opened 2018-09-04
Support of gzipped CSV data sources

chlewe:
Closes #37. For a detailed description of the new features, refer to the corresponding issue.

PR #51, opened 2018-10-09
Added support for N-Triples and gzipped N-Triples

chlewe:
Closes #45. _(For a more detailed description of the new features, please refer to the corresponding issue.)_

- Added `RdfFileDataSource`.
- Tested `RdfFileDataSource` on the _vlog java-bindings_ level, on the _vlog4j_ level, as well as on the __vlog4j_level_.
- Wrote an example that uses `.nt.gz` files.
- Did some ~minor~ code refactoring.


vlog4j-examples/src/main/java/org/semanticweb/vlog4j/examples/core/AddDataFromRdfFile.java
```diff
@@ -0,0 +1,170 @@
...
+ */
+public class AddDataFromRdfFile {
+
+	public static void main(final String[] args)
+			throws EdbIdbSeparationException, IOException, ReasonerStateException, IncompatiblePredicateArityException {
+
+		/* 1. Instantiating entities and rules. */
+		final Predicate triplesEDB = Expressions.makePredicate("triplesEDB", 3);
+		final Predicate triplesIDB = Expressions.makePredicate("triplesIDB", 3);
+
```
irina-dragoste: The encoding of the Bicycle example into a program with triples is correct. However, the standard way of handling unary predicates in RDF (ex: bycicle, wheel) is with special predicate "rdf:type" (the long URI being <http://www.w3.org/1999/02/22-rdf-syntax-ns#type>).
For example, to specify that constant <c> is a bicycle, you have triple
<http://an.example/c > < http://www.w3.org/1999/02/22-rdf-syntax-ns#type > < http://an.example/bicycle > .

This way, you don't need the emptyObject, and the atoms in facts and rules become more intuitive, as they respect standard subject ("c")-predicate("has type")-object("bycicle") format. 

Note that VLog reasoner does not have a special interpretation for < http://www.w3.org/1999/02/22-rdf-syntax-ns#type >, and it is treated as any other predicate during reasoning.

vlog4j-examples/src/main/java/org/semanticweb/vlog4j/examples/core/AddDataFromRdfFile.java
```diff
@@ -0,0 +1,170 @@
...
+		final Rule rule7 = Expressions.makeRule(isPartOfIDB, hasPartIDBReversed);
+
+		/*
+		 * triplesIDB(?s, <~/hasPartIDB>, ?o) :- triplesIDB(?o, <~/isPartOfIDB>, ?s) .
+		 */
+		final Atom isPartOfIDBReversed = Expressions.makeAtom(triplesIDB, o, isPartOfPredicate, s);
+		final Rule rule8 = Expressions.makeRule(hasPartIDB, isPartOfIDBReversed);
+
+		/* 2. Loading, reasoning, and querying. */
+		final Reasoner reasoner = Reasoner.getInstance();
```
irina-dragoste: in examples project, it is a good idea to show how our system can be used most advantageously. Therefore, we should use try with resources whenever possible (perhaps except in one examples class, to display an alternative, and notify user to call reasoner.close()). 
Also, we can statically import Expressions.makeConstant, Expressions.makeVariable, Expressions.makeAtom etc, to obtain less lengthy code.

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/reasoner/implementation/RdfFileDataSourceTest.java
```diff
@@ -0,0 +1,67 @@
...
+		assertEquals("file", rdfFileDataSource.getFileNameWithoutExtension());
+	}
+
+	@Test(expected = NullPointerException.class)
+	public void fileNameNotNull() throws IOException {
+		new CsvFileDataSource(null);
+	}
+
+	@Test(expected = IllegalArgumentException.class)
+	public void fileNameEndsWithCsv() throws IOException {
```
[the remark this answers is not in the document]
chlewe: It seems as though an outdated version of `RdfFileDataSourceTest` made it into the pull request. Thank you for noticing that! I will push an update soon.

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/reasoner/implementation/LoadDataFromRdfTest.java
```diff
@@ -0,0 +1,134 @@
...
+		assertFalse(unexistingFile.exists());
+		final FileDataSource fileDataSource = new RdfFileDataSource(unexistingFile);
+
+		try (final Reasoner reasoner = Reasoner.getInstance()) {
+			reasoner.addFactsFromDataSource(Expressions.makePredicate(ternaryPredicateName, 3), fileDataSource);
+			reasoner.load();
+		}
+	}
+
+}
```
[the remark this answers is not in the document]
chlewe: I see the need for this test. However, there is a problem: Because N-Triples enforces a predicate arity of 3 by its syntax, `toVLogPredicate()` returns a predicate of arity -1 on files with arities other than 3 (erroneous syntax). Therefore, `validateDataSourcePredicateArities()`only outputs a warning and never throws an `IncompatiblePredicateArityException`, because of the `if`-`else` structure. I can add the test, but it will fail until we add new behaviour.

chlewe:
Please check the code again. All issues should have been resolved. The only remaining problem is the test for `IncompatiblePredicateArityException` with `RdfFileDataSource`, which is currently failing and therefore ignored.

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/reasoner/implementation/LoadDataFromRdfFileTest.java
```diff
@@ -0,0 +1,116 @@
...
+		assertFalse(nonexistingFile.exists());
+		final FileDataSource fileDataSource = new RdfFileDataSource(nonexistingFile);
+
+		try (final Reasoner reasoner = Reasoner.getInstance()) {
+			reasoner.addFactsFromDataSource(ternaryPredicate, fileDataSource);
+			reasoner.load();
+		}
+	}
+
+	@Ignore
```
irina-dragoste: I see, we need to discuss with the VLog people on how to handle this problem.

vlog4j-examples/src/main/java/org/semanticweb/vlog4j/examples/core/AddDataFromCsvFile.java
```diff
@@ -0,0 +1,167 @@
...
+/**
+ * This example shows how facts can be imported from {@code .csv.gz} files, and
+ * also how query answers can be exported to {@code .csv} files. <i>Note that
+ * you can also import from {@code .csv} files.</i>
+ *
+ * For importing, a {@link CsvFileDataSource} that contains a path to the
+ * corresponding {@code .csv.gz} file must be created. A {@code .csv} file
+ * contains facts over exactly one predicate in the CSV format, while a
+ * {@code .csv.gz} file is the gzipped version of such a {@code .csv} file. For
+ * exporting, a path to the output {@code .csv} file must be specified.
```
irina-dragoste: Before "For exporting..." I would add, in a new paragraph: "Also, facts resulting from answering a query can be exported into .csv files. "

vlog4j-examples/src/main/java/org/semanticweb/vlog4j/examples/core/AddDataFromRdfFile.java
```diff
@@ -0,0 +1,167 @@
...
+import org.semanticweb.vlog4j.core.reasoner.DataSource;
+import org.semanticweb.vlog4j.core.reasoner.Reasoner;
+import org.semanticweb.vlog4j.core.reasoner.exceptions.EdbIdbSeparationException;
+import org.semanticweb.vlog4j.core.reasoner.exceptions.IncompatiblePredicateArityException;
+import org.semanticweb.vlog4j.core.reasoner.exceptions.ReasonerStateException;
+import org.semanticweb.vlog4j.core.reasoner.implementation.RdfFileDataSource;
+import org.semanticweb.vlog4j.examples.ExamplesUtils;
+
+/**
+ * This example is an adaptation of {@link AddDataFromCsvFile}. It shows how
```
irina-dragoste: "This example is an adaptation of {}.." I would given more details. You can specify that this example shows how to import facts from files of N-Triples format. Then, after the details about valid extensions, you can say that the rules and facts in this example have been adapted from the csv example to the N-Triples format.

vlog4j-examples/src/main/java/org/semanticweb/vlog4j/examples/core/AddDataFromRdfFile.java
```diff
@@ -0,0 +1,167 @@
...
+		final Atom hasPartIDB = makeAtom(triplesIDB, s, hasPartPredicate, o);
+		final Atom hasPartEDB = makeAtom(triplesEDB, s, hasPartPredicate, o);
+		final Rule rule2 = makeRule(hasPartIDB, hasPartEDB);
+
+		/*
+		 * triplesIDB(?s, <~/isPartOf>, ?o) :- triplesEDB(?s, <~/isPartOf>, ?o) .
+		 */
+		final Atom isPartOfIDB = makeAtom(triplesIDB, s, isPartOfPredicate, o);
+		final Atom isPartOfEDB = makeAtom(triplesEDB, s, isPartOfPredicate, o);
+		final Rule rule3 = makeRule(isPartOfIDB, isPartOfEDB);
```
irina-dragoste: As the middle term of these ternary predicates is seen by the reasoner as a simple term (it does not have the interpretation of a predicate), you can have a single rule instead of rule1, rule2 and rule3:

final Variable p = makeVariable("p");
	/*
		 * We will write '<~/someName>' instead of <http://an.example/someName> and
		 * <~#someName> instead of <www.w3.org/1999/02/22-rdf-syntax-ns#someName>.
		 *
		 * triplesIDB(?s,  ?p, ?o) :- triplesEDB(?s, ?p, ?o) .
		 */
		final Atom atomIDB = makeAtom(triplesIDB, s, p, o);
		final Atom atomEDB = makeAtom(triplesEDB, s, p, o);
		final Rule rule1 = makeRule(atomIDB , atomEDB );

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/reasoner/implementation/LoadDataFromRdfFileTest.java
irina-dragoste: Please add a javadoc to explain why this test is ignored, and a TODO (for example, decide on the expected behavior of VLog)

irina-dragoste: I have discussed with the developers of VLog. The parser ignores lines of an .nt file which do not have a valid N-Triples format. Therefore, the given test file seems to contain no facts over a given ternary predicate (both lines are not N-Triple). This is the expected behavior. I guess this unit test can be modified to test this, instead?

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/reasoner/FileDataSourceUtils.java
```diff
@@ -68,4 +84,60 @@ private FileDataSourceUtils() {
...
 
+	public static void testConstructor(final FileDataSource unzippedFileDataSource,
+			final FileDataSource zippedFileDataSource)
+					throws IOException {
+		final String expectedDirCanonicalPath = new File(INPUT_FOLDER).getCanonicalPath();
+		final File unzippedFile = unzippedFileDataSource.getFile();
+		final File zippedFile = zippedFileDataSource.getFile();
+
+		assertEquals(unzippedFile, unzippedFileDataSource.getFile());
+		assertEquals(expectedDirCanonicalPath, unzippedFileDataSource.getDirCanonicalPath());
```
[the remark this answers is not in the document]
chlewe: I see what you mean. There is, however, a huge problem: What would those expected parameters look like? At the moment, the parameters that the function takes are abstract `FileDataSource` objects and not paths. One of the major ideas when writing these tests was to create the concrete `CsvFileDataSource` or `RdfFileDataSource` objects outside of these functions and to only work with the generalised `FileDataSource` in the utility class. If I passed paths to the utility class, I would have to pass the concrete class as parameter as well and create the concrete objects with a generic constructor, which is very troublesome and ugly in Java. Moreover, as far as I understand, constructors with parameters are not supported in that approach, which is not compatible with `CsvFileDataSource` or `RdfFileDataSource`, because they expect a path.

vlog4j-examples/src/main/java/org/semanticweb/vlog4j/examples/core/AddDataFromCsvFile.java
chlewe: This information is already given in the first paragraph of this example. _"This example shows [...] how query answers can be exported to .csv files."_ I can reformat the Javadoc a bit, but I wouldn't change any content.

chlewe:
All issues should be resolved now. Please re-check the code.

vlog4j-examples/src/main/java/org/semanticweb/vlog4j/examples/core/AddDataFromRdfFile.java
```diff
@@ -0,0 +1,160 @@
...
+import org.semanticweb.vlog4j.core.reasoner.DataSource;
+import org.semanticweb.vlog4j.core.reasoner.Reasoner;
+import org.semanticweb.vlog4j.core.reasoner.exceptions.EdbIdbSeparationException;
+import org.semanticweb.vlog4j.core.reasoner.exceptions.IncompatiblePredicateArityException;
+import org.semanticweb.vlog4j.core.reasoner.exceptions.ReasonerStateException;
+import org.semanticweb.vlog4j.core.reasoner.implementation.RdfFileDataSource;
+import org.semanticweb.vlog4j.examples.ExamplesUtils;
+
+/**
+ * This example is an adaptation of {@link AddDataFromCsvFile}, where the rules
```
irina-dragoste: For more clarity, I would start with the sentence "This example shows facts can be imported from files in RDF N-Triples format".

irina-dragoste: For more clarity, I would start with the sentence "This example shows facts can be imported from files in RDF N-Triples format".

vlog4j-examples/src/main/java/org/semanticweb/vlog4j/examples/core/AddDataFromRdfFile.java
```diff
@@ -0,0 +1,160 @@
...
+ * have been modified to work with the ternary predicates that N-Triples
+ * enforces. It shows how facts can be imported from {@code .nt.gz} files.
+ * <i>(Note that you can also import from {@code .nt} files.)</i> Moreover, this
+ * example shows how query answers can be exported to {@code .csv} files.
+ *
+ * For importing, an {@link RdfFileDataSource} that contains a path to the
+ * corresponding {@code .nt.gz} file must be created. An {@code .nt} file
+ * contains facts over one or more predicates in the N-Triples format, while an
```
irina-dragoste: "facts over one or more predicates" is confusing (we use the term Predicate to refer to the predicate of the fact fed to the reasoner, that we associate with the terms in the file, and not the middle term in the <subject> <predicate> <object>.

chlewe:
Sorry. I seem to have been lost in all the comments here, which led me to overlooking one. It should be fixed now. I also removed the loops.

vlog4j-examples/src/main/java/org/semanticweb/vlog4j/examples/core/AddDataFromRdfFile.java
```diff
@@ -44,17 +44,21 @@
...
- * example shows how query answers can be exported to {@code .csv} files.
- *
+ * enforces.
+ * <p>
  * For importing, an {@link RdfFileDataSource} that contains a path to the
  * corresponding {@code .nt.gz} file must be created. An {@code .nt} file
- * contains facts over one or more predicates in the N-Triples format, while an
- * {@code .nt.gz} file is the gzipped version of such an {@code .nt} file.
- *
+ * contains facts exactly one ternary predicate in the RDF N-Triples format,
```
irina-dragoste: missing adverb: "contains facts exactly one ternary predicate".  I would still say "contains facts in N-Triples format that can be associated to a ternary {@link Predicate}", since the predicate is not in the file, but associated to it via reasoner#addFactsFromDataSource().

vlog4j-examples/src/main/java/org/semanticweb/vlog4j/examples/core/AddDataFromRdfFile.java
```diff
@@ -44,17 +44,21 @@
 import org.semanticweb.vlog4j.examples.ExamplesUtils;
 
 /**
+ * This example shows how facts can be imported from {@code .nt.gz} files.
```
irina-dragoste: Since the example classes are most probably the first lines of code a user would read, I suggest mentioning the N-Triples format explicitly at the end of the Javadoc, with a link to the online W3C specification (like in RDFFileDataSource). Indeed, it is inconsistent with the ".csv" example javadoc, but we believe it is important to mention this format here, besides just the ".nt" extension.

chlewe:
I adapted the Javadoc of the RDF example and test accordingly and even made the CSV example and test consistent with these changes.

Issue #53, opened 2018-10-10
Core: add unit tests for Reasoner#setReasoningTimeout


irina-dragoste:

irina-dragoste:
test cases when:
* reasoning timeout is null
* reasoning timeout is Integer.MAX_INT
* materialisations terminates before timeout, test the output of Reasoner#reason()

irina-dragoste:
@Adrian-Bielefeldt did you already start working on issue #53 ? If so, please assign yourself, and commit your code on a separate branch, making a pull request. If not, please let me know, I will take over the issue.

Adrian-Bielefeldt:
I started working on this, but I do not seem to be able to assign myself to this issue, since I do not have write access to this repository.

PR #58, opened 2018-11-25
53: Added unit test for reasoner timeout

shade-belisar:
Added a unit test to check the reasoner timeout as per [issue 53](https://github.com/knowsys/vlog4j/issues/53).
The test checks both skolem and restricted chase using a small set of
existential rules generating infinite data.
A small overhead of one second is accepted to allow for setup and tear
down of reasoner resources.

### DIFF
--- a/vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/reasoner/ReasonerTimeoutTest.java
+++ b/vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/reasoner/ReasonerTimeoutTest.java
@@ -56,7 +56,13 @@ public class ReasonerTimeoutTest {
 	 */
 	private static int timeout = 1;
 	
+	/**
+	 * A list of facts to be used in multiple test runs.
+	 */
 	private static List<Atom> facts = new ArrayList<>();
+	/**
+	 * A list of rules to be used in multiple test runs.
+	 */
 	private static List<Rule> rules = new ArrayList<>();
 	
 	private Reasoner reasoner;
@@ -67,6 +73,15 @@ public class ReasonerTimeoutTest {
 	@org.junit.Rule
 	public Timeout globalTimeout = Timeout.seconds(timeout + 1);
 
+	/**
+	 * This method provides the {@link #facts} and {@link #rules} to be used in all test runs.
+	 * To test if the timeout works as expected, a small set of facts and rules is used that results in an infinite chase.
+	 * Facts:
+	 * 	infinite_EDB(A, B)
+	 * Rules:
+	 * 	infinite_IDB(?x, ?y) :- infinite_EDB(?x, ?y) 
+	 * 	infinite_IDB(?y, ?z) :- infinite_IDB(?x, ?y)
+	 */
 	@BeforeClass
 	public static void setUpBeforeClass() {
 		Predicate infinite_EDB = makePredicate("infinite_EDB", 2);
@@ -85,8 +100,8 @@ public class ReasonerTimeoutTest {
 		
 		Variable z = makeVariable("z");
 		
-		Atom further_yz = makeAtom(infinite_IDB, y, z);
-		Rule infinite_rule = makeRule(further_yz, infinite_IDB_xy);
+		Atom infinite_IDB_yz = makeAtom(infinite_IDB, y, z);
+		Rule infinite_rule = makeRule(infinite_IDB_yz, infinite_IDB_xy);
 		rules.add(infinite_rule);
 	}
 


vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/reasoner/ReasonerTimeoutTest.java
```diff
@@ -0,0 +1,99 @@
...
+	
+	/**
+	 * The timeout after which reasoning should be completed in seconds.
+	 */
+	static int timeout = 1;
+	
+	static List<Atom> facts = new ArrayList<>();
+	static List<Rule> rules = new ArrayList<>();
+	
+	Reasoner reasoner;
```
irina-dragoste: please make this state variable private (even though it is a unit test).

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/reasoner/ReasonerTimeoutTest.java
```diff
@@ -0,0 +1,99 @@
...
+ * Results are accepted within one second to account for setup and tear down of reasoning resources.
+ * @author Adrian Bielefeldt
+ *
+ */
+public class ReasonerTimeoutTest {
+	
+	/**
+	 * The timeout after which reasoning should be completed in seconds.
+	 */
+	static int timeout = 1;
```
irina-dragoste: what is the reason for which these fields are package protected? I would make fields private until we need to use them in other classes.

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/reasoner/ReasonerTimeoutTest.java
```diff
@@ -0,0 +1,99 @@
...
+	Reasoner reasoner;
+	
+	/**
+	 * The timeout after which reasoning should be completed. One second is added to account for setup and tear down of reasoning resources.
+	 */
+	@org.junit.Rule
+	public Timeout globalTimeout = Timeout.seconds(timeout + 1);
+
+	@BeforeClass
+	public static void setUpBeforeClass() throws Exception {
```
irina-dragoste: Even though it is mentioned in the class javadoc, I think it is best to document this method, and specify that the set of rules and facts it constructs leads to an infinite (skolem or restricted) chase (has an infinite universal model).

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/reasoner/ReasonerTimeoutTest.java
```diff
@@ -0,0 +1,99 @@
...
+	
+	@Test
+	public void restricted() throws EdbIdbSeparationException, IncompatiblePredicateArityException, IOException, ReasonerStateException {
+		reasoner.setAlgorithm(Algorithm.RESTRICTED_CHASE);
+		
+		reasoner.load();
+		
+		reasoner.reason();
+	}
+}
```
irina-dragoste: Please add an @After method, in which you call reasoner.close(), in order to discard the C++ resource. It is not a good practice to leave that to the garbage collector. That is why Reasoner implements AutoCloseable, but a try with resources cannot be used in the way you designed the test.

irina-dragoste: I would also add two test methods in which setTimeout() will be called after calling load(), as the load() method on the Java side corresponds to creating the reasoner resource on the C++ side, and the timeout information could have been lost.

irina-dragoste: Before commit, please run maven clean install, as this generates the license text in the header of each class.

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/reasoner/ReasonerTimeoutTest.java
```diff
@@ -0,0 +1,99 @@
...
+	static List<Atom> facts = new ArrayList<>();
+	static List<Rule> rules = new ArrayList<>();
+	
+	Reasoner reasoner;
+	
+	/**
+	 * The timeout after which reasoning should be completed. One second is added to account for setup and tear down of reasoning resources.
+	 */
+	@org.junit.Rule
+	public Timeout globalTimeout = Timeout.seconds(timeout + 1);
```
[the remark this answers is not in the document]
Adrian-Bielefeldt: Rule fields must be public and not static according to the JUnit documentation.

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/reasoner/ReasonerTimeoutTest.java
Adrian-Bielefeldt: All class variables are now private.

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/reasoner/ReasonerTimeoutTest.java
Adrian-Bielefeldt: All class variables are now private.

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/reasoner/ReasonerTimeoutTest.java
Adrian-Bielefeldt: Added method javadoc explaining that and how an infinite chase is constructed.

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/reasoner/ReasonerTimeoutTest.java
Adrian-Bielefeldt: Closing of resources is added.

Adrian-Bielefeldt: Added two additional test methods that set the timeout after the reasoner was loaded.

Adrian-Bielefeldt: The license text has been added.

PR #65, opened 2019-03-20
Don't try to use a non-existing compiler

mmarx:
Only force gcc-5 if it is actually available.


mmarx:
Please note that the CI checks need to be rerun after clearing the build cache, as this change is not covered by the tests otherwise.

mmarx:
Still builds with a clean cache.

PR #78, opened 2019-08-15
New rule syntax

mkroetzsch:
Add support for parsing rules from our own new syntax, supporting negation, existentials, and RDF-style terms.


mkroetzsch:
The DOID example and simple example currently do not work as intended, since there are no more queries. It would be good if parsing single literals would work, so that we can create the queries without builiding objects in Java. Alternatively, we should at least do the latter.

PR #88, opened 2019-09-12
Client

larryjgonzalez:
New module to create a stand-alone executable jar file for VLog4j


mkroetzsch:
I suppose this code still needs some Maven configurations to build the jar, right? Here is an example of how I did this in the past:

* [WDTK client module pom](https://github.com/Wikidata/Wikidata-Toolkit/blob/v0.8.0/wdtk-client/pom.xml); note the packaging ```jar```
* [WDTK distribution module](https://github.com/Wikidata/Wikidata-Toolkit/tree/v0.8.0/wdtk-distribution); this "virtual" module was used to keep the build configuration for the client (not sure why this cannot be in the client pom; note that this also builds library jars for the other modules, which we do not really need, so maybe one could simplify this; feel free to do it differently if there are other options).
* [WDTK parent pom](https://github.com/Wikidata/Wikidata-Toolkit/blob/v0.8.0/pom.xml); the corresponding parent pom, not sure if we need anything from there

If everything is set up correctly, it should be possible to build the jar as follows:
```
  change into "wdtk-distribution" directory
  mvn -Pclient install
  mvn -Plibrary install
  sign jars manually:
     change into "target" directory
     gpg --detach-sig wdtk-client-0.5.0.jar
```

larryjgonzalez:
I see. Many thanks.
I just exported an executable jar from Eclipse. I'll look into these Maven configurations.

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,161 @@
...
+				System.out.println("  --query: " + queries.get(queries.size()-1).toString());
+			} catch (ParsingException e) {
+				System.out.println("Failed to parse query: " + queryString.toString() + ". " + e.getMessage());
+				System.exit(1);
+			}
+		}
+
+		/* Print configuration */
+		System.out.println("  --log-file: " + logFile);
+		System.out.println("  --log-level: " + logLevel.toString());
```
irina-dragoste: same as above, you don't need to call toString()

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,161 @@
...
+		System.out.println("  --chase-algorithm: " + chaseAlgorithm.toString());
```
irina-dragoste: same as above, you don't need to call toString()

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,161 @@
...
+		System.out.println("  --timeout: " + ((timeout > 0) ? Integer.toString(timeout) : "none"));
```
irina-dragoste: same as above, you don't need to call toString()

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,161 @@
...
+//				System.out.println("Saving model ...");
+//			}
+
+			if (queries.size() > 0) {
+				System.out.println("Answering queries ...");
+				for (PositiveLiteral query : queries) {
+					if (SaveQueryResult.saveQueryResults) {
+						try {
+							reasoner.exportQueryAnswersToCsv(query,
+									SaveQueryResult.outputQueryResultFolder +"/"+ query.toString()+".csv", true);
```
irina-dragoste: save `SaveQueryResult.outputQueryResultFolder +"/"+ query.toString()+".csv"` into a local variable.

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,161 @@
...
+						} catch (IOException e) {
+							System.out.println("Cant save query " + query.toString());
```
irina-dragoste: Typo ("Cant")

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientTestAcyclicity.java
```diff
@@ -0,0 +1,19 @@
+package org.vlog4j.client.picocli;
+
+import picocli.CommandLine.Command;
+
+@Command(name = "testacyclicity", description = "Test if the rule set satisfy any acyclicity notion")
```
irina-dragoste: Typo ('the rule set satisfy" should be "the rule set satisfies")

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientTestAcyclicity.java
```diff
@@ -0,0 +1,19 @@
...
+public class VLog4jClientTestAcyclicity implements Runnable {
+
+//	@Option(names = "--acyclicity-notion", required = false, description = "Acyclicity notion. One of:JA (Joint Acyclicity), RJA (Restricted Joint Acyclicity), RFA (Model-Faithful Acyclicity), RMFA (Restricted Model-Faithful Acyclicity). All by default.")
```
irina-dragoste: add a TODO

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,161 @@
...
+				System.out.println("Something went wrong. Please check the file paths inside the rule files.");
+				e.printStackTrace();
+			}
+
+//			// TODO save the model
+//			if (saveModel.saveModel) {
+//				System.out.println("Saving model ...");
+//			}
+
+			if (queries.size() > 0) {
```
irina-dragoste: if the IO exception actually does occur above, the code below will still be reached.
I don't think we want to answer queries if there was an IO exception, or do we?

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,161 @@
...
+		for (String ruleFile : ruleFiles) {
+			try {
+				RuleParser.parseInto(kb, new FileInputStream(ruleFile));
+				System.out.println("  --rule-file: " + ruleFile);
+			} catch (FileNotFoundException e) {
+				System.out.println("File not found: " + ruleFile + ". " + e.getMessage());
+			} catch (ParsingException e) {
+				System.out.println("Failed to parse rule file: " + ruleFile + ". " + e.getMessage());
+			}
+		}
```
irina-dragoste: Do we want to continue reasoning if we encounter an exception while reading rules?

larryjgonzalez:
I've implemented the requested changes. Could you please check them?

vlog4j-core/pom.xml
```diff
@@ -60,5 +61,54 @@
 				</plugins>
 			</build>
 		</profile>
+		<profile>
+			<id>client</id>
+			<dependencies>
+				<dependency>
+					<groupId>${project.groupId}</groupId>
+					<artifactId>vlog4j-client</artifactId>
```
irina-dragoste: I need to look into this, it seems that we have a cyclic dependency. Let's discuss soon.

vlog4j-client/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+			<artifactId>vlog4j-examples</artifactId>
+			<version>${project.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>info.picocli</groupId>
+			<artifactId>picocli</artifactId>
+			<version>${picoli.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>fr.lirmm.graphik</groupId>
```
irina-dragoste: is graal-api used? if not, the dependency should be removed.

vlog4j-client/src/main/java/org/vlog4j/client/picocli/PrintQueryResults.java
```diff
@@ -0,0 +1,66 @@
...
+		} else {
+			return true;
+		}
+	}
+
+	/**
+	 * Print configuration error and exit the program
+	 */
+	public void printErrorAndExit() {
+		System.out.println("Configuration error: --print-query-result-size and "
```
irina-dragoste: why do you append two strings, instead of having only one string?

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,187 @@
...
+	// @ArgGroup(exclusive = false)
+	// private SaveModel saveModel = new SaveModel();
+
+	@Override
+	public void run() {
+		ExamplesUtils.configureLogging();
+
+		if (!printQueryResults.isConfigOk()) {
+			printQueryResults.printErrorAndExit();
+		}
```
irina-dragoste: you can extract these lines into a method
`printQueryResults.validate()`

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,187 @@
...
+
+		if (!saveQueryResults.isConfigOk()) {
+			saveQueryResults.printErrorAndExit();
+		}
```
irina-dragoste: same as above

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,187 @@
...
+		System.out.println("Configuration:");
+
+		/* Configure rules */
+		final KnowledgeBase kb = new KnowledgeBase();
+		for (String ruleFile : ruleFiles) {
+			try {
+				RuleParser.parseInto(kb, new FileInputStream(ruleFile));
+				System.out.println("  --rule-file: " + ruleFile);
+			} catch (FileNotFoundException e) {
+				System.out.println("File not found: " + ruleFile + ". " + e.getMessage());
```
irina-dragoste: Write error to System.err

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,187 @@
...
+		// if (!saveModel.isConfigOk()) {
+		// saveModel.printErrorAndExit();
+		// }
+
+		System.out.println("Configuration:");
+
+		/* Configure rules */
+		final KnowledgeBase kb = new KnowledgeBase();
+		for (String ruleFile : ruleFiles) {
+			try {
```
irina-dragoste: the best practice is to have only the statements that generate exceptions surrounded by try-catch. Also, you can separate the two statements that throw two different exceptions here on two lines.

```
final FileInputStream fileInputStream;
try{fileInputStream = FileInputStream(ruleFile);
} catch(FileNotFoundException  e){
...
}
try{RuleParser.parseInto(kb, fileInputStream);
} catch ....
```

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,187 @@
...
+				RuleParser.parseInto(kb, new FileInputStream(ruleFile));
+				System.out.println("  --rule-file: " + ruleFile);
+			} catch (FileNotFoundException e) {
+				System.out.println("File not found: " + ruleFile + ". " + e.getMessage());
+				System.exit(1);
+			} catch (ParsingException e) {
+				System.out.println("Failed to parse rule file: " + ruleFile + ". " + e.getMessage());
+				System.exit(1);
```
irina-dragoste: Write error to System.err

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,187 @@
...
+
+		/* Configure queries (We throw parsing query errors before materialization) */
+		List<PositiveLiteral> queries = new ArrayList<>();
+		for (String queryString : queryStrings) {
+			try {
+				queries.add(RuleParser.parsePositiveLiteral(queryString));
+				System.out.println("  --query: " + queries.get(queries.size() - 1));
+			} catch (ParsingException e) {
+				System.out.println("Failed to parse query: " + queryString + ". " + e.getMessage());
+				System.exit(1);
```
irina-dragoste: You could not exit, but print a message and continue with the other queries.

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,187 @@
...
+		}
+
+		/* Configure queries (We throw parsing query errors before materialization) */
+		List<PositiveLiteral> queries = new ArrayList<>();
+		for (String queryString : queryStrings) {
+			try {
+				queries.add(RuleParser.parsePositiveLiteral(queryString));
+				System.out.println("  --query: " + queries.get(queries.size() - 1));
+			} catch (ParsingException e) {
+				System.out.println("Failed to parse query: " + queryString + ". " + e.getMessage());
```
irina-dragoste: Write error to System.err

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,187 @@
...
+			reasoner.setLogLevel(logLevel);
+			// chaseAlgorithm
+			reasoner.setAlgorithm(chaseAlgorithm);
+			// timeout
+			if (timeout > 0) {
+				reasoner.setReasoningTimeout(timeout);
+			}
+
+			try {
+				System.out.println("Executing the chase ...");
```
irina-dragoste: remove this line from the try block, the try blocks should contain only what could cause the exception.

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,187 @@
...
+				reasoner.reason();
+			} catch (IOException e) {
+				System.out.println("Something went wrong. Please check the log file.");
```
irina-dragoste: Write error to System.err

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,187 @@
...
+				System.exit(1);
+			}
+
+			// TODO save the model
+			// if (saveModel.saveModel) {
+			// System.out.println("Saving model ...");
+			// }
+
+			String outputPath;
+			if (queries.size() > 0) {
```
irina-dragoste: with Collections, ` .isEmpty()` method is used, instead of comparing size to 0.

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,187 @@
...
+				System.out.println("Answering queries ...");
+				for (PositiveLiteral query : queries) {
+					if (saveQueryResults.saveResults) {
+						try {
+							outputPath = saveQueryResults.outputQueryResultDirectory + "/" + query + ".csv";
```
irina-dragoste: remove statement that does not cause exception from try block

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,187 @@
...
+							reasoner.exportQueryAnswersToCsv(query, outputPath, true);
+						} catch (IOException e) {
+							System.out.println("Can't save query " + query);
```
irina-dragoste: Write error to System.err

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,187 @@
...
+							System.exit(1);
```
irina-dragoste: Instead of exiting, we should write a message and continue with the other queries.

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,187 @@
...
+						}
+					}
+					if (printQueryResults.sizeOnly) {
+						System.out.println(
+								"Elements in " + query + ": " + ExamplesUtils.getQueryAnswerCount(query, reasoner));
```
irina-dragoste: Number of query answers.

Why call query answers "elements"?

vlog4j-client/src/test/java/org/vlog4j/client/PrintQueryResultsTest.java
```diff
@@ -0,0 +1,63 @@
+package org.vlog4j.client;
```
irina-dragoste: why not have the same package name for test classes as for source classes?
Then, you can declare some methods package-protected and test them here, while making sure they cannot be accessed from elsewhere.

vlog4j-client/src/test/java/org/vlog4j/client/PrintQueryResultsTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+import org.junit.Assert;
+
+import org.junit.Test;
+import org.vlog4j.client.picocli.PrintQueryResults;
+
+public class PrintQueryResultsTest {
+
+	@Test
+	public void sizeOnly() {
```
irina-dragoste: These unit tests only test configurations. Rename method names or test class name accordingly.

pom.xml
```diff
@@ -67,6 +67,7 @@
 		<owlapi.version>5.1.11</owlapi.version>
 		<openrdf.sesame.version>2.7.16</openrdf.sesame.version>
 		<graal.version>1.3.1</graal.version>
+		<picoli.version>4.0.2</picoli.version>
```
irina-dragoste: Also, why does the parent pom not include the vlog4j-client module?

vlog4j-core/pom.xml
irina-dragoste: This does not even work with mvn install -Pclient. 
Why is this profile not declared in vlog4j-client, (or in parent), but in vlog4j-core?

vlog4j-core/pom.xml
```diff
@@ -60,5 +61,54 @@
...
+					<version>${project.version}</version>
+				</dependency>
+			</dependencies>
+			<build>
+				<plugins>
+					<plugin>
+						<groupId>org.apache.maven.plugins</groupId>
+						<artifactId>maven-shade-plugin</artifactId>
+						<executions>
```
irina-dragoste: There is no version declared for maven-shade-plugin

pom.xml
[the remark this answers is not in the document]
larryjgonzalez: The latest version was 4.0.2 when I started the branch. I'm updating it to 4.0.4 now.

vlog4j-client/src/main/java/org/vlog4j/client/picocli/PrintQueryResults.java
```diff
@@ -0,0 +1,66 @@
...
+	 * If true, Vlog4jClient will print the query result in stdout. Mutually
+	 * exclusive with {@code --print-query-result-size}
+	 * 
+	 * @default false
+	 */
+	@Option(names = "--print-complete-query-result", description = "Boolean. If true, Vlog4jClient will print the query result in stdout. False by default.")
+	public boolean complete = false;
+
+	public boolean isConfigOk() {
+		if (sizeOnly & complete) {
```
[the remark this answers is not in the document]
larryjgonzalez: Thank you. This is indeed much more readable. I'm implementing it

vlog4j-client/src/main/java/org/vlog4j/client/picocli/PrintQueryResults.java
```diff
@@ -0,0 +1,66 @@
...
+public class PrintQueryResults {
+
+	/**
+	 * If true, Vlog4jClient will print the size of the query result. Mutually
+	 * exclusive with {@code --print-complete-query-result}
+	 * 
+	 * @default true
+	 */
+	@Option(names = "--print-query-result-size", description = "Boolean. If true, Vlog4jClient will print the size of the query result. True by default.")
+	public boolean sizeOnly = true;
```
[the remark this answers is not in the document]
larryjgonzalez: I've made them private and added setters and getters.

vlog4j-client/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+			<artifactId>vlog4j-rdf</artifactId>
+			<version>${project.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>${project.groupId}</groupId>
+			<artifactId>vlog4j-parser</artifactId>
+			<version>${project.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>${project.groupId}</groupId>
```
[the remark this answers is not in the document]
larryjgonzalez: Now it only uses `vlog4j-core`, `vlog4j-parser`, and `vlog4j-examples` (to print query answers). In the future it should use the rest of the modules. I'll remove the unused dependencies, and add them when necessary.

vlog4j-client/src/main/java/org/vlog4j/client/picocli/SaveModel.java
```diff
@@ -0,0 +1,77 @@
...
+
+import picocli.CommandLine.Option;
+
+public class SaveModel {
+
+	/**
+	 * If true, Vlog4jClient will save the model in {@code --output-model-directory}
+	 *
+	 * @default false
+	 */
```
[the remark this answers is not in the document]
larryjgonzalez: I've made them private and added setters and getters.

vlog4j-client/src/main/java/org/vlog4j/client/picocli/PrintQueryResults.java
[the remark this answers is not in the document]
larryjgonzalez: Thank you. I was not aware of this feature

vlog4j-client/src/main/java/org/vlog4j/client/picocli/SaveModel.java
```diff
@@ -0,0 +1,77 @@
...
+	/**
+	 * Directory to store the model. Used only if {@code --store-model} is true.
+	 *
+	 * @default "model"
+	 */
+	@Option(names = "--output-model-directory", description = "Directory to store the model. Used only if --store-model is true. \"model\" by default.")
+	public String outputModelDirectory = "model";
+
+	public boolean isConfigOk() {
+		if (saveModel & (outputModelDirectory == null || outputModelDirectory.isEmpty())) {
```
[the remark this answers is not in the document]
larryjgonzalez: Thank you. I've used `return !saveModel || (outputModelDirectory != null && !outputModelDirectory.isEmpty());`

vlog4j-client/src/main/java/org/vlog4j/client/picocli/SaveQueryResults.java
```diff
@@ -0,0 +1,80 @@
...
+ * limitations under the License.
+ * #L%
+ */
+
+import java.io.File;
+
+import picocli.CommandLine.Option;
+
+public class SaveQueryResults {
+
```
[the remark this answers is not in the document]
larryjgonzalez: I've made them private, and I've added setters and getters

vlog4j-client/src/main/java/org/vlog4j/client/picocli/SaveQueryResults.java
```diff
@@ -0,0 +1,80 @@
...
+	 * Directory to store the model. Used only if {@code --save-query-results} is
+	 * true
+	 *
+	 * @default query-results
+	 */
+	@Option(names = "--output-query-result-directory", description = "Directory to store the model. Used only if --save-query-results is true. \"query-results\" by default.")
+	public String outputQueryResultDirectory = "query-results";
+
+	public boolean isConfigOk() {
+		if (saveResults & (outputQueryResultDirectory == null || outputQueryResultDirectory.isEmpty())) {
```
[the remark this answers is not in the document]
larryjgonzalez: same answer as well, :)

vlog4j-client/src/main/java/org/vlog4j/client/picocli/PrintQueryResults.java
```diff
@@ -0,0 +1,66 @@
...
+	/**
+	 * If true, Vlog4jClient will print the query result in stdout. Mutually
+	 * exclusive with {@code --print-query-result-size}
+	 * 
+	 * @default false
+	 */
+	@Option(names = "--print-complete-query-result", description = "Boolean. If true, Vlog4jClient will print the query result in stdout. False by default.")
+	public boolean complete = false;
+
+	public boolean isConfigOk() {
```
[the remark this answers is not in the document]
larryjgonzalez: It is more informative. Thank you

vlog4j-client/src/main/java/org/vlog4j/client/picocli/SaveModel.java
```diff
@@ -0,0 +1,77 @@
...
+
+	/**
+	 * Directory to store the model. Used only if {@code --store-model} is true.
+	 *
+	 * @default "model"
+	 */
+	@Option(names = "--output-model-directory", description = "Directory to store the model. Used only if --store-model is true. \"model\" by default.")
+	public String outputModelDirectory = "model";
+
+	public boolean isConfigOk() {
```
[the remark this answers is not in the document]
larryjgonzalez: It is more informative. Thank you

vlog4j-client/src/main/java/org/vlog4j/client/picocli/SaveQueryResults.java
```diff
@@ -0,0 +1,80 @@
...
+	/**
+	 * Directory to store the model. Used only if {@code --save-query-results} is
+	 * true
+	 *
+	 * @default query-results
+	 */
+	@Option(names = "--output-query-result-directory", description = "Directory to store the model. Used only if --save-query-results is true. \"query-results\" by default.")
+	public String outputQueryResultDirectory = "query-results";
+
+	public boolean isConfigOk() {
```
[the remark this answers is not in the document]
larryjgonzalez: It is more informative. Thank you

vlog4j-core/pom.xml
larryjgonzalez: This is strange. I think I don't have a cyclic dependency in the code. Also, mvn install -Pclient work for me. Yes, let's discuss about it soon, :)
I had placed `client` in beside `vlog-core` because `development` was there. I've moved it to parent.

vlog4j-client/pom.xml
larryjgonzalez: It is not required now. I've removed it, but I'll add it when necessary.

vlog4j-client/src/main/java/org/vlog4j/client/picocli/PrintQueryResults.java
larryjgonzalez: To avoid horizontal scrolling

vlog4j-client/src/main/java/org/vlog4j/client/picocli/SaveModel.java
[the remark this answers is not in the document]
larryjgonzalez: Thanks, :)

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
larryjgonzalez: It is fixed now. Thanks

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientTestAcyclicity.java
larryjgonzalez: It is fixed now. Thanks

vlog4j-client/src/main/java/org/vlog4j/client/picocli/PrintQueryResults.java
```diff
@@ -0,0 +1,66 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+
+import picocli.CommandLine.Option;
+
+public class PrintQueryResults {
```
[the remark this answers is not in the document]
larryjgonzalez: Javadoc added

vlog4j-client/src/main/java/org/vlog4j/client/picocli/SaveModel.java
```diff
@@ -0,0 +1,77 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+
+import java.io.File;
+
+import picocli.CommandLine.Option;
+
```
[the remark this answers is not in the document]
larryjgonzalez: Javadoc added

vlog4j-client/src/main/java/org/vlog4j/client/picocli/SaveQueryResults.java
```diff
@@ -0,0 +1,80 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+
+import java.io.File;
+
+import picocli.CommandLine.Option;
+
```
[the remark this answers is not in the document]
larryjgonzalez: Javadoc added

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,187 @@
...
+
+import picocli.CommandLine.ArgGroup;
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Option;
+
+@Command(name = "materialize", description = "Execute the chase and store the literal's extensions")
+public class VLog4jClientMaterialize implements Runnable {
+
+	// TODO add link to rls syntax
+	@Option(names = "--rule-file", description = "Rule file(s) in rls syntax", required = true)
```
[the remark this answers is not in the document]
larryjgonzalez: RLS syntax is our new syntax. I'll add a link to our [wiki](https://github.com/knowsys/vlog4j/wiki/RLS-syntax) when it is ready

vlog4j-client/src/test/java/org/vlog4j/client/PrintQueryResultsTest.java
larryjgonzalez: Thank you. This was a silly mistake

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
irina-dragoste: actually, a method would be better.

vlog4j-client/pom.xml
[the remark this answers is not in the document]
irina-dragoste: Since vlog4j-examples has dependencies on almost every module, we can create a lightweight vlog4j-utils module in another issue, and move there the methods for printing query answers.

create vlog4j-utils module #124

vlog4j-client/src/main/java/org/vlog4j/client/picocli/SaveQueryResults.java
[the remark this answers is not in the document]
irina-dragoste: Perhaps add details like in what format the query results will be saved, and where.
The same for the javadoc of SaveModel.

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
larryjgonzalez: Thank you. It is indeed much more readable

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -0,0 +1,187 @@
...
+				System.exit(1);
+			}
+		}
+
+		/* Configure queries (We throw parsing query errors before materialization) */
+		List<PositiveLiteral> queries = new ArrayList<>();
+		for (String queryString : queryStrings) {
+			try {
+				queries.add(RuleParser.parsePositiveLiteral(queryString));
+				System.out.println("  --query: " + queries.get(queries.size() - 1));
```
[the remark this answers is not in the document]
larryjgonzalez: Thank you

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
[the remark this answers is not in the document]
larryjgonzalez: Thank you

vlog4j-client/src/main/java/org/vlog4j/client/picocli/PrintQueryResults.java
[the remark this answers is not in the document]
larryjgonzalez: Thank you

vlog4j-client/src/main/java/org/vlog4j/client/picocli/SaveModel.java
```diff
@@ -0,0 +1,77 @@
...
+		} else {
+			return true;
+		}
+	}
+
+	/**
+	 * Print configuration error and exit the program
+	 */
+	public void printErrorAndExit() {
+		System.out.println("Configuration error: --save-model requires a non-null --output-model-directory.");
```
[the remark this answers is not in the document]
larryjgonzalez: Thank you

vlog4j-client/src/main/java/org/vlog4j/client/picocli/SaveQueryResults.java
```diff
@@ -0,0 +1,80 @@
...
+		} else {
+			return true;
+		}
+	}
+
+	/**
+	 * Print configuration error and exit the program
+	 */
+	public void printErrorAndExit() {
+		System.out.println(
```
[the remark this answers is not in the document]
larryjgonzalez: Thank you

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
larryjgonzalez: Thank you. I've created the method and updated the tests accordingly

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
larryjgonzalez: Thank you

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
larryjgonzalez: Maybe I am missing something here, but I don't think this would work because the variable `fileInputStream` won't be defined in scope of the second try.

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
larryjgonzalez: Thank you

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
larryjgonzalez: Thank you

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
larryjgonzalez: Thank you

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
larryjgonzalez: Thank you. This was a silly mistake

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
larryjgonzalez: Thank you

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
larryjgonzalez: Thank you

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
larryjgonzalez: This was solved when we moved that functionality into the `doSaveQueryResults` method

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
larryjgonzalez: Thank you

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
larryjgonzalez: Thank you

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
larryjgonzalez: Thank you

vlog4j-client/src/main/java/org/vlog4j/client/picocli/VLog4jClientMaterialize.java
irina-dragoste: You are right, you would have to initialise the variable, because the compiler does not know that if the first exception is thrown, then the program will terminate due to System.exit.
Then the code would look like below, which is not very nice, either. You can keep your version, but move the System.out out of the try{} block. 

```
for (final String ruleFile : this.ruleFiles) {
			FileInputStream fis = null;
			try {
				fis = new FileInputStream(ruleFile);

			} catch (final FileNotFoundException e) {
				System.err.println("File not found: " + ruleFile + ". " + e.getMessage());
				System.exit(1);
			}
			try {
				RuleParser.parseInto(kb, fis);
			} catch (final ParsingException e) {
				System.err.println("Failed to parse rule file: " + ruleFile + ". " + e.getMessage());
				System.exit(1);
			}
			System.out.println("  --rule-file: " + ruleFile);
		}
```

pom.xml
irina-dragoste: As discussed, the parent pom needs to include the vlog4j-client module.

PR #89, opened 2019-09-12
Remove static imports

larryjgonzalez:


mkroetzsch:
This seems to need some update to merge.

larryjgonzalez:
True.
After merging #90 and #91 with master, this branch (#89 ) was not in conditions to be merged.
Creating a new pull request now.

vlog4j-parser/src/main/java/org/semanticweb/vlog4j/parser/RuleParser.java
```diff
@@ -70,6 +71,16 @@ public static KnowledgeBase parse(String input) throws ParsingException {
 		return parse(inputStream, "UTF-8");
 	}
 
+	public static Rule parseRule(String input) throws ParsingException {
+		InputStream inputStream = new ByteArrayInputStream(input.getBytes());
+		JavaCCParser localParser = new JavaCCParser(inputStream, "UTF-8");
+		try {
+			return localParser.rule();
+		} catch (ParseException | PrefixDeclarationException | TokenMgrError e) {
+			throw new ParsingException(e.getMessage(), e);
```
irina-dragoste: An error should be logged before throwing the exception. Also, the exception message should contain more info.
If you pass the exception e as a parameter to the ParsingException, you don't need to propagate e.getMessage(): it will be automatically propagated (see below). Instead, you should write an explicit message like "Error during rule parsing".


The following program outputs the lines below.

`public static void main(final String[] args) throws Exception {
		try {
			throw new Exception("a");
		} catch (final Exception e) {
			throw new Exception("b", e);
		}
	}`

Exception in thread "main" java.lang.Exception: b
	at playground.Main.main(Main.java:9)
Caused by: java.lang.Exception: a
	at playground.Main.main(Main.java:7)

irina-dragoste: please implement these changes in all relevant methods in this class.

vlog4j-examples/src/main/java/org/semanticweb/vlog4j/examples/core/ConfigureReasonerLogging.java
```diff
@@ -76,23 +68,18 @@
...
+		List<Rule> rules = Arrays.asList(
+				/* A(?x, ?y) :- A_EDB(?x, ?y) . */
+				RuleParser.parseRule("A(?X, ?Y) :- A_EDB(?X, ?Y) ."),
+				/* exists z. B(?y, !z) :- A(?x, ?y) . */
+				RuleParser.parseRule("B(?Y, !Z) :- A(?X, ?Y) ."),
+				/* B(?y, ?x), A(?y, ?x) :- B(?x, ?y) . */
+				RuleParser.parseRule("B(?Y, ?X), A(?Y, ?X) :- B(?X, ?Y) ."));
+
+		/* A(c,d) */
+		Fact fact = (Fact) RuleParser.parsePositiveLiteral("A_EDB(\"c\",\"d\")");
```
irina-dragoste: This casting generates an exception.

vlog4j-examples/src/main/java/org/semanticweb/vlog4j/examples/core/ConfigureReasonerLogging.java
```diff
@@ -76,23 +68,18 @@
...
-
-	/* A(c,d) */
-	private static Fact fact = Expressions.makeFact("A_EDB", Arrays.asList(makeConstant("c"), makeConstant("d")));
-
-	public static void main(final String[] args) throws IOException {
+	public static void main(final String[] args) throws IOException, ParsingException {
+
+		List<Rule> rules = Arrays.asList(
+				/* A(?x, ?y) :- A_EDB(?x, ?y) . */
+				RuleParser.parseRule("A(?X, ?Y) :- A_EDB(?X, ?Y) ."),
```
irina-dragoste: This results in a parsing exception.

larryjgonzalez:
Thank @irina-dragoste for your valuable help and comments.
I've implemented the method RuleParser.parseFact, and update ConfigureReasonerLogging example to avoid the exception. You can find the changes online this branch

PR #92, opened 2019-09-12
Complement of transitive closure example

larryjgonzalez:
Add new simple example


irina-dragoste:
we do not want to have this feature

Issue #84, opened 2019-08-28
support successive reasoning without re-loading datasources


irina-dragoste:

irina-dragoste:
Currently, when reason() is called successively, the reasoner is reset, and all datasources have to be loaded again.

Issue #96, opened 2019-09-16
refactor unit tests to a smaller, clear scope


irina-dragoste:

larryjgonzalez:
The [Java Programming Style Guidelines published by Petroware](https://petroware.no/javastyle.html) that we follow does not have any guideline about tests.

Before refactoring unit test, I would suggest to decide which convention we are going to follow. As an example, which unit test naming convention we are going to use.

PR #119, opened 2019-10-07
Emulate iterat reas removal #84

irina-dragoste:
Pull request for issue #84.
This implementation sets QA correctness to INCORRECT for any KB change. The correctness can be further refined (ex: when there were no rules in the initial KB and only fact addition has been performed, when the same statements have been added and removed, etc)


vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/KnowledgeBase.java
```diff
@@ -220,6 +256,74 @@ public void addStatements(Statement... statements) {
...
+	 * @return true, if the knowledge base has changed.
+	 */
+	boolean doRemoveStatement(Statement statement) {
+		Validate.notNull(statement, "Statement cannot be Null!");
+
+		if (this.statements.contains(statement) && statement.accept(this.removeStatementVisitor)) {
+			this.statements.remove(statement);
+			return true;
+		}
+		return false;
```
larryjgonzalez: Should we use `else { return false;  }` ?

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/KnowledgeBase.java
```diff
@@ -220,6 +256,74 @@ public void addStatements(Statement... statements) {
...
+	}
+
+	/**
+	 * Removes a collection of statements to the knowledge base.
+	 * 
+	 * @param statements
```
larryjgonzalez: Should we describe the parameter, instead of just name it?

CerielJacobs:
I would be more concerned about the javadoc, which says that the statement is added, not removed :-)

Best,	Ceriel

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/KnowledgeBase.java
irina-dragoste: I don't think it's necessary.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/KnowledgeBase.java
irina-dragoste: done

irina-dragoste:
> I would be more concerned about the javadoc, which says that the statement is added, not removed :-) Best, Ceriel

Thank you for catching this,
done.

larryjgonzalez:
* I've used the auto-formatter from eclipse in several files.
* I don't have suggestions for changes in classes under `src/main` directory
* I have several suggestions for the classes under `src/test` directory. But they should be addressed in #96

PR #125, opened 2019-10-21
Remove parse fact function

larryjgonzalez:
Use RuleParser.parseInto instead of RuleParser.parseFact
Remove RuleParser.parseFact


alloka:
I think this is ok, because I don't find any use cases for parse fact. but what about the parserule ?

vlog4j-examples/src/main/java/org/semanticweb/vlog4j/examples/core/ConfigureReasonerLogging.java
```diff
@@ -69,11 +69,11 @@ public static void main(final String[] args) throws IOException, ParsingExceptio
 		try (final Reasoner reasoner = Reasoner.getInstance()) {
 			final KnowledgeBase kb = reasoner.getKnowledgeBase();
 			/* exists z. B(?y, !z) :- A(?x, ?y) . */
-			kb.addStatements(RuleParser.parseRule("B(?Y, !Z) :- A(?X, ?Y) ."));
```
alloka: this is parserule function, not parse fact

vlog4j-examples/src/main/java/org/semanticweb/vlog4j/examples/core/ConfigureReasonerLogging.java
```diff
@@ -69,11 +69,11 @@ public static void main(final String[] args) throws IOException, ParsingExceptio
...
+			RuleParser.parseInto(kb, "B(?Y, !Z) :- A(?X, ?Y) .");
 			/* B(?y, ?x), A(?y, ?x) :- B(?x, ?y) . */
-			kb.addStatements(RuleParser.parseRule("B(?Y, ?X), A(?Y, ?X) :- B(?X, ?Y) ."));
+			RuleParser.parseInto(kb, "B(?Y, ?X), A(?Y, ?X) :- B(?X, ?Y) .");
```
alloka: this one as well, I don't know, maybe the parse rule is not important as well and we can just rely on parseinto

vlog4j-examples/src/main/java/org/semanticweb/vlog4j/examples/core/ConfigureReasonerLogging.java
larryjgonzalez: True. While the goal of the branch was to remove the parseFact method, I also updated how we are parsing rules into a KB.

alloka: did you remove parserule ?

larryjgonzalez: The method `parseRule` is still part of the class `RuleParser`. Here I updated how we are adding some rules into a `KnowledgeBase`.

Users are still able to use `RuleParser.parseRule` in order to get an instance of `Rule`

vlog4j-examples/src/main/java/org/semanticweb/vlog4j/examples/core/ConfigureReasonerLogging.java
larryjgonzalez: `RuleParser.parseRule` is still part of the code.

larryjgonzalez:
Following previous conversation, I'm closing this pull request.

Issue #86, opened 2019-08-29
Core: update toString() in core.model entities 


irina-dragoste:
The toString() method in Rules and other core.model entities shows a different representation than our syntax in the parser module.
Have separate data objects for existential and universal variables.

## Task description
For every String s, if s is parsable and parse(s) = an entity o, then o.toString() should return s.
Make sure s is escaped to be parsable.

## Specifications
Note that for constants and predicates, there are several possible string representations that get parsed into the same String.

### DatatypeConstants
For DatatypeConstants, there are some abbreviated parsable String representations (see https://www.w3.org/TeamSubmission/turtle/ ). The String representation of these constants should be the abbreviated one.

### Predicates and AbstractConstants

We distinguish between **absoulte** IRIs (that contain ":" character) and **relative** IRIs.
The String representation of a relative IRI does not require to be surrounded by angle brackets ("<" and ">"), while the String representation of an absoulte IRI  does.

Note that Predicate names represent absolute or relative IRIs.

Note that IRI atom terms are parsed into `AbstractConstant`s. Therefore, an `AbstractConstant` represents an IRI.

Note some AbstractConstant relative IRIs may look like the (long or abbreviated) String representation of a datatype constant. (ex: \"something\", 123, 0.5, 1e ). In this case, to ensure roundtripping, the String representation of these relative IRI AbstractConstants requires angle brackets.

## Testing  "round-tripping"
We should test "round-tripping", that the string used to parse a data model object `o` is the same as the one returned by `o.toString()` (or that if we parse `o.toString()`, we obtain an object equal to `o`).

Test for Predicates: tests for absolute and relative IRIs

  Test for Abstract Constants: tests for absolute and relative IRIs,
test for AbstractConstants with name=the String representation of a DatatypeConstant of each type.

  Test for Datatype Constants (strings that are parsed into objects of type DatatypeConstant
		
		1. generic test for custom datatype
		example datatypeConstantLongString="\"a\"^^<whatever>"; //ex: Literal literal = RuleParser.parseLiteral("P(\"a\"^^<whatever>)"); 
		test that datatypeConstantLongString == parseResult(datatypeConstantLongString).toString()
		
				test that other datatypes (ex: float, boolean etc) behave like above
		
		2. test for XSD_String datatype
		example datatypeConstantStringShortString="\"a\""; // ex: Literal literal = RuleParser.parseLiteral("P(\"a\")"); 
[...]

PR #130, opened 2019-10-27
Add serializer class with toString methods #86

alloka:
please check if this needs something


mkroetzsch:
Let me clarify the task a bit. The new class should be the place where all the serialisation code is. The toString() methods of the implementations instead should call the methods of the static class to create strings.

mkroetzsch:
P.S. Also, all new code should be tested with suitable unit tests. For this branch, the tests should be in the parser module and test for successful round-tripping.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,98 @@
...
+			} else {
+				stringBuilder.append(", ");
+			}
+			stringBuilder.append(term);
+		}
+		stringBuilder.append(")");
+		return stringBuilder.toString();
+	}
+
+	public static String getConstantString(AbstractConstantImpl constant) {
```
irina-dragoste: `AbstractConstant `param

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,98 @@
...
+		return constant.getName();
+	}
+
+	public static String getExistentialVarString(ExistentialVariableImpl existentialvariable) {
```
irina-dragoste: `ExistentialVariable `param

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,98 @@
...
+		return "!" + existentialvariable.getName();
+	}
+
+	public static String getUniversalVarString(UniversalVariableImpl universalvariable) {
```
irina-dragoste: `UniversalVariable `param

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,98 @@
...
+		return "?" + universalvariable.getName();
+	}
+
+	public static String getDatatypeConstantString(DatatypeConstantImpl datatypeconstant) {
```
irina-dragoste: `DatatypeConstant `param

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,98 @@
...
+		return datatypeconstant.getName();
+	}
+
+	public static String getNamedNullString(NamedNullImpl namednull) {
```
irina-dragoste: `NamedNull `param

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,98 @@
...
+		return "_" + namednull.getName();
+	}
+
+	public static String getLanguageConstantString(LanguageStringConstantImpl languagestringconstant) {
```
irina-dragoste: `LanguageStringConstant `param

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/DataSourceDeclarationImpl.java
```diff
@@ -85,7 +85,7 @@ public boolean equals(final Object obj) {
 
 	@Override
 	public String toString() {
-		return "@source " + this.predicate.toString() + "(" + this.predicate.getArity() + ") : "
+		return "@source " + Serializer.getPredicateString(this.predicate) + "(" + this.predicate.getArity() + ") : "
```
irina-dragoste: This does not correspond to our syntax.
`getPredicateString` returns also the arity of the predicate.
it should be
`"@source " + this.predicate.getName() + "(" + this.predicate.getArity() + ") : " + this.dataSource.toConfigString() + " .";`

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/model/ConjunctionImplTest.java
```diff
@@ -181,4 +181,19 @@ public void negativeLiteralsNoNullElements() {
...
+		final Variable y = Expressions.makeUniversalVariable("Y");
+		final Constant c = Expressions.makeAbstractConstant("c");
+		final Constant d = Expressions.makeAbstractConstant("d");
+		final PositiveLiteral positiveLiteral1 = Expressions.makePositiveLiteral("p", x, c);
+		final PositiveLiteral positiveLiteral2 = Expressions.makePositiveLiteral("p", y, x);
+		final PositiveLiteral positiveLiteral3 = Expressions.makePositiveLiteral("q", x, d);
+		final List<PositiveLiteral> positiveLiteralList = Arrays.asList(positiveLiteral1, positiveLiteral2,
+				positiveLiteral3);
+		final Conjunction<PositiveLiteral> conjunction1 = new ConjunctionImpl<>(positiveLiteralList);
+		assertEquals("p(?X, c), p(?Y, ?X), q(?X, d)", conjunction1.toString());
```
irina-dragoste: It would be more interesting to test a conjunction of PositiveLiterals, NegativeLiterals, and Facts

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/model/DataSourceDeclarationTest.java
```diff
@@ -63,4 +62,16 @@ public void equalityTest() throws MalformedURLException {
...
+	public void testdataSourceDecalarationtoString() throws MalformedURLException {
+		DataSource dataSource1 = new SparqlQueryResultDataSource(new URL("https://example.org/"), "var",
+				"?var wdt:P31 wd:Q5 .");
+		Predicate predicate1 = Expressions.makePredicate("p", 3);
+		DataSourceDeclaration dataSourceDeclaration1 = new DataSourceDeclarationImpl(predicate1, dataSource1);
+		DataSource dataSource2 = new SparqlQueryResultDataSource(new URL("https://example.org/"), "var",
+				"?var wdt:P31 wd:Q5 .");
+		Predicate predicate2 = Expressions.makePredicate("p", 3);
+		DataSourceDeclaration dataSourceDeclaration2 = new DataSourceDeclarationImpl(predicate2, dataSource2);
+		assertEquals(dataSourceDeclaration1.toString(), dataSourceDeclaration2.toString());
```
irina-dragoste: this method does not really test toString(), but the constructors. You test that by creating two objects with the same parameters, you get the same (string representations of the )objects.

You should test what you expect to get from toString() instead, by comparing it to a String constant.

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/model/ConjunctionImplTest.java
```diff
@@ -181,4 +181,19 @@ public void negativeLiteralsNoNullElements() {
 		Expressions.makeConjunction(negativeLiteralList);
 	}
 
+	@Test
+	public void testconjunctiontoString() {
```
irina-dragoste: careful with the method names. There are various popular naming conventions (https://dzone.com/articles/7-popular-unit-test-naming), but all of them require proper camelCase.

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/model/DataSourceDeclarationTest.java
```diff
@@ -63,4 +62,16 @@ public void equalityTest() throws MalformedURLException {
 		assertFalse(dataSourceDeclaration1.equals(null)); // written like this for recording coverage properly
 	}
 
+	@Test
+	public void testdataSourceDecalarationtoString() throws MalformedURLException {
```
irina-dragoste: same as above for method names.

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/model/FactTest.java
```diff
@@ -54,4 +54,13 @@ public void factsOnlyContainConstants() {
 		new FactImpl(p, Arrays.asList(x));
 	}
 
+	@Test
+	public void testtoString() {
```
irina-dragoste: same as above for method names.

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/model/RuleImplTest.java
```diff
@@ -137,4 +138,21 @@ public void noUnsafeVariables() {
 		Expressions.makeRule(literal1, literal2);
 	}
 
+	@Test
+	public void testtoString() {
```
irina-dragoste: same as above for method names.

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/model/RuleImplTest.java
```diff
@@ -137,4 +138,21 @@ public void noUnsafeVariables() {
...
+		final Variable x = Expressions.makeUniversalVariable("X");
+		final Variable y = Expressions.makeExistentialVariable("Y");
+		final Variable z = Expressions.makeUniversalVariable("Z");
+		final Constant c = Expressions.makeAbstractConstant("c");
+		final PositiveLiteral atom1 = Expressions.makePositiveLiteral("p", x, c);
```
irina-dragoste: I would include different types of Literals and Constants in this test.

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/model/TermImplTest.java
```diff
@@ -132,6 +132,42 @@ public void namedNullGetterTest() {
 		assertEquals(TermType.NAMED_NULL, n.getType());
 	}
 
+	@Test
```
irina-dragoste: careful with method names.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/FactImpl.java
```diff
@@ -47,5 +47,11 @@ public FactImpl(Predicate predicate, List<Term> terms) {
 	public <T> T accept(StatementVisitor<T> statementVisitor) {
 		return statementVisitor.visit(this);
 	}
+	
+	@Override
+	public String toString() {
```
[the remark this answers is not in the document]
irina-dragoste: I would include the "." inside getLiteralString()

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/model/TermImplTest.java
alloka: This is not my test, I did not implement this.

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/model/RuleImplTest.java
alloka: I guess This is not mine as well, I added tests for different type of constants.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/FactImpl.java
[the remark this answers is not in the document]
alloka: but we don't want to do that for all literals.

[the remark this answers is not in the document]
irina-dragoste: ok

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/ConjunctionImpl.java
```diff
@@ -95,7 +95,7 @@ public String toString() {
 			} else {
```
irina-dragoste: why don't you call getConjunctionString() inside the toString() method?
currently, toString() does the same thing as getConjunctionString(), so the code is duplicated.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,124 @@
...
+		}
+		stringBuilder.append(")");
+		return stringBuilder.toString();
+	}
+
+	public static String getConstantString(Constant constant) {
+		return constant.getName();
+	}
+
+	public static String getExistentialVarString(ExistentialVariable existentialvariable) {
```
irina-dragoste: use camel case for variable names.
(i.e. eachWordStartsWithUppercaseExceptTheFirstWord)

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,124 @@
...
+		return "!" + existentialvariable.getName();
+	}
+
+	public static String getUniversalVarString(UniversalVariable universalvariable) {
```
irina-dragoste: use camel case for variable names.
(i.e. eachWordStartsWithUppercaseExceptTheFirstWord)

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,124 @@
...
+		return "?" + universalvariable.getName();
+	}
+
+	public static String getDatatypeConstantString(DatatypeConstant datatypeconstant) {
```
irina-dragoste: use camel case for variable names.
(i.e. eachWordStartsWithUppercaseExceptTheFirstWord)

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,124 @@
...
+		return datatypeconstant.getName();
+	}
+
+	public static String getNamedNullString(NamedNull namednull) {
```
irina-dragoste: use camel case for variable names.
(i.e. eachWordStartsWithUppercaseExceptTheFirstWord)

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,124 @@
...
+
+	public static String getLanguageConstantString(LanguageStringConstant languagestringconstant) {
+		return languagestringconstant.getName();
+	}
+
+	public static String getPredicateString(Predicate predicate) {
+		return " Predicate [ name= " + predicate.getName() + ", arity= " + predicate.getArity() + "]";
+	}
+
+	public static String getDataSourceDeclarationString(DataSourceDeclaration datasourcedeclaration) {
```
irina-dragoste: use camel case for variable names.
(i.e. eachWordStartsWithUppercaseExceptTheFirstWord)

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,124 @@
...
+		return "@source " + getPredicateString(datasourcedeclaration.getPredicate()) + "("
+				+ datasourcedeclaration.getPredicate().getArity() + ") : "
```
irina-dragoste: This does not return the same syntax as the parser.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/ConjunctionImpl.java
alloka: Ok, I thought I already removed it

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
alloka: can you give me the syntax that is returned by the parser because I don't know how to obtain it for Datasource delarations

alloka: ok, I've seen the syntax, but let's discuss this in the meeting

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,116 @@
...
+ *
+ */
+public final class Serializer {
+
+	private Serializer() {
+
+	}
+
+	public static String getRuleString(Rule rule) {
+		return rule.getHead() + " :- " + rule.getBody() + ".";
```
irina-dragoste: this calls `toString() `on the head and body `Conjunctions`, which in turn calls `getConjunctionString()`. We should not depend to the toString method in `Conjunction`, but only on the `Serializer` methods. That is why this method should return
`getConjunctionString(rule.getHead())+" :- " + getConjunctionString(rule.getBody())+"."`

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,116 @@
...
+		}
+		stringBuilder.append(literal.getPredicate().getName()).append("(");
+		boolean first = true;
+		for (final Term term : literal.getArguments()) {
+			if (first) {
+				first = false;
+			} else {
+				stringBuilder.append(", ");
+			}
+			stringBuilder.append(term);
```
irina-dragoste: this also calls toString() on the term. We need to discuss if this is how we want this method to be implemented.

alloka: Yes, we can discuss this in the meeting, because we don't want to repeat the variables and constants methods.

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/model/DataSourceDeclarationTest.java
alloka: ok, you are right, but I need first to know about datasource Declarations, maybe someone can explain them to me in the meeting.

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/model/RuleImplTest.java
alloka: the rule calls the tostring of every literal and constant type it has, so if you have a test for all the types anywhere else that works, be sure that it will appear the same anywhere else.

alloka:
@mkroetzsch  @irina-dragoste  I guess now the whole thing is ready for review

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/NamedNullImpl.java
```diff
@@ -46,8 +46,12 @@ public NamedNullImpl(final String name) {
 		return termVisitor.visit(this);
 	}
 
+	public String getSyntacticRepresentation() {
```
irina-dragoste: add override annotation, move to NamedNull

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/PredicateImpl.java
```diff
@@ -87,9 +85,13 @@ public boolean equals(Object obj) {
 		return this.arity == other.getArity() && this.name.equals(other.getName());
 	}
 
+	public String getSyntacticRepresentation() {
```
irina-dragoste: add override annotation, move to Predicate

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/RuleImpl.java
```diff
@@ -103,9 +103,14 @@ public boolean equals(final Object obj) {
 		return this.head.equals(other.getHead()) && this.body.equals(other.getBody());
 	}
 
+	
+	public String getSyntacticRepresentation() {
```
irina-dragoste: add override annotation, move to Rule

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/UniversalVariableImpl.java
```diff
@@ -44,8 +44,12 @@ public UniversalVariableImpl(final String name) {
 		return termVisitor.visit(this);
 	}
 
+	public String getSyntacticRepresentation() {
```
irina-dragoste: add override annotation, move to UniversalVariable

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/InMemoryDataSource.java
```diff
@@ -101,4 +101,10 @@ public String toConfigString() {
 		return null;
 	}
 
+	@Override
+	public String getSyntacticRepresentation() {
+		// TODO Auto-generated method stub
```
irina-dragoste: This method should not return null.
@mkroetzsch what should we return?
We don't want to display data if it is too big.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/CsvFileDataSource.java
```diff
@@ -73,4 +73,9 @@ public String toString() {
 		return "CsvFileDataSource [csvFile=" + getFile() + "]";
 	}
 
+	@Override
+	public String getSyntacticRepresentation() {
+		return "load-csv(\"" + getFile() + "\") .";
+	}
+
```
irina-dragoste: don't we want the toString() of DataSources to return getSyntacticRepresentation()?
@mkroetzsch what do you think?

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -129,6 +130,15 @@ public boolean equals(Object obj) {
 			return predicate.equals(other.predicate);
 		}
 
+		public String getSyntacticRepresentation() {
```
irina-dragoste: we don't want a syntactic representation for a reasoner. Btw, how does this even compile? There is no getString(Reasoner) in the Serializer.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,143 @@
+package org.semanticweb.vlog4j.core.model.implementation;
+
+import java.util.List;
```
irina-dragoste: remove unused imports. You can see them as warnings, in yellow in Eclipse. Please fix all warnings.

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/model/DataSourceDeclarationTest.java
```diff
@@ -63,4 +70,38 @@ public void equalityTest() throws MalformedURLException {
...
+	public void dataSourceDecalarationToStringTest() throws MalformedURLException {
+		DataSource dataSource1 = new SparqlQueryResultDataSource(new URL("https://example.org/"), "var",
+				"?var wdt:P31 wd:Q5 .");
+		Predicate predicate1 = Expressions.makePredicate("p", 3);
+		DataSourceDeclaration dataSourceDeclaration1 = new DataSourceDeclarationImpl(predicate1, dataSource1);
+		DataSource dataSource2 = new SparqlQueryResultDataSource(new URL("https://example.org/"), "var",
+				"?var wdt:P31 wd:Q5 .");
+		Predicate predicate2 = Expressions.makePredicate("p", 3);
+		DataSourceDeclaration dataSourceDeclaration2 = new DataSourceDeclarationImpl(predicate2, dataSource2);
+		assertEquals(dataSourceDeclaration1.toString(), dataSourceDeclaration2.toString());
```
irina-dragoste: please fix old comment

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/EntityTest.java
```diff
@@ -0,0 +1,152 @@
...
+	final List<Literal> LiteralList = Arrays.asList(positiveLiteral1, positiveLiteral2, positiveLiteral3,
+			NegativeLiteral, PositiveLiteral4);
+	final Conjunction<Literal> bodyLiterals = Expressions.makeConjunction(atom1, atom2);
+	final Conjunction<PositiveLiteral> headPositiveLiterals = Expressions.makePositiveConjunction(headAtom1);
+	final Conjunction<Literal> bodyConjunction = new ConjunctionImpl<>(LiteralList);
+	final Rule rule1 = new RuleImpl(headPositiveLiterals, bodyLiterals);
+	final Rule rule2 = new RuleImpl(headPositiveLiterals, bodyConjunction);
+
+	@Test
+	public void factToStringRoundTripTest() {
```
irina-dragoste: this is not round tripping.
 
You are supposed to parse an entity e from a String s, then test that e.toString() (or e.getSyntacticRepresentation()) is equal to s.

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/EntityTest.java
```diff
@@ -0,0 +1,152 @@
...
+		KnowledgeBase kb = new KnowledgeBase();
+		KnowledgeBase kb2 = new KnowledgeBase();
+		kb.addStatement(f1);
```
irina-dragoste: why do you add things to the knowledge base?

alloka: because I don't want to give a certain string I wanted to create the object then parse and then I can check whether the string is the same as the one before the parsing

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/EntityTest.java
alloka: are you sure?, because I think that can work too.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
alloka: because it has a local class that implements DataSourceDeclaration and I don't know what to do about it because DataSourceDeclaration extends entity

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/CsvFileDataSource.java
alloka: I was told by Markus to leave this as it is.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,143 @@
...
+				stringBuilder.append(", ");
+			}
+			stringBuilder.append(term.getSyntacticRepresentation());
+		}
+		stringBuilder.append(")");
+		return stringBuilder.toString();
+	}
+
+	public static String getString(Fact fact) {
+		final StringBuilder stringBuilder = new StringBuilder("");
```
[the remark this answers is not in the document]
alloka: yes, we need the "." but when i do this solution it gives a stackoverflow exception because I guess that it thinks that it is trying to call itself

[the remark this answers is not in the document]
alloka: but its ok I changed its name

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,143 @@
...
+	public static String getString(UniversalVariable universalVariable) {
+		return "?" + universalVariable.getName();
+	}
+
+	public static String getString(NamedNull namedNull) {
+		return "_" + namedNull.getName();
+	}
+
+	public static String getString(Predicate predicate) {
+		return " Predicate [ name= " + predicate.getName() + ", arity= " + predicate.getArity() + "]";
```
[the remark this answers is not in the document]
mmarx: The “usual” logic programming notation would be `predicate/arity`, but even `predicate(arity)` is already better than `name = predicate, arity = arity`.

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/EntityTest.java
```diff
@@ -0,0 +1,124 @@
...
+	@Test
+	public void ruleToStringRoundTripTest() throws ParsingException {
+		assertEquals(RuleParser.parseRule(rule1.toString()), RuleParser.parseRule("q(?X, !Z) :- p(?X, c), p(?X, ?Y)."));
+		assertEquals(RuleParser.parseRule(rule2.toString()),
+				RuleParser.parseRule("q(?X, !Z) :- p(?X, c), p(?Y, ?X), q(?X, d), ~r(?X, d), s(c, \"Test\"@en)."));
+	}
+
+	@Test
+	public void dataSourceDeclarationToStringParsingTest() throws ParsingException, IOException {
+		KnowledgeBase kb = new KnowledgeBase();
```
irina-dragoste: I would simplify the unit tests, they are quite difficult to follow now:
- test only one object per test method
- compare a string with the toString on the object that results from parsing that string

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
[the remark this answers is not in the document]
irina-dragoste: I would have implemented it as below, but I think your solution is better, as it avoids casting.
```
public static String getString(final Fact fact) {
		return getString((Literal) fact) + ".";
	}
```

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/EntityTest.java
alloka: The only way to test this is through a rule, shouldn't be easier to create a rule with all different types and test them, instead of just creating new rule for every type as it is pointless to do so ?

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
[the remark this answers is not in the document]
alloka: I guess we need to discuss this with markus in the meeting, as I don't know anything about this or what to do with it

alloka:
> > I guess this checks if it is parsable and at the same time if it is the desired string
> 
> This is overcomplicated in my opinion, and difficult to read.
> Why not have
> 
> ```
> String factString="some string";
> Fact fact=RuleParser.parseFact(factString);
> assertEquals(factString,fact.toString());
> ```

I guess mine is better, because I parse both the string and the fact that I created and check if they are the same? I remember being told by markus at some point that I had to parse both, maybe we can discuss this as well in the meeting

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -42,15 +42,31 @@
...
-
+	/**
+	 * Constructor.
+	 */
 	private Serializer() {
 
 	}
 
+	/**
+	 * Creates a String representation of a given {@link Rule}. Example: "p(?X) :-
```
irina-dragoste: I would explain what the example represents, perhaps link to getString(Literal literal), and would include existential variables in our example. We should also link to our syntax wiki.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -42,15 +42,31 @@
...
+	 * @param rule a {@link Rule}
+	 * @return String representation corresponding to a given {@link Rule}.
+	 */
 	public static String getString(Rule rule) {
 		return getString(rule.getHead()) + " :- " + getString(rule.getBody()) + ".";
 	}
 
+	/**
+	 * Creates a String representation of a given {@link Literal}. Example:
+	 * "~q(?X,?Y)"
```
irina-dragoste: Please explain that this is an example of a NegativeLiteral, and show and example of a PositiveLiteral.

Also, include constants in the example.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -70,36 +86,92 @@ public static String getString(Literal literal) {
...
+	 * 
+	 * @param fact a {@link Fact}
+	 * @return String representation corresponding to a given {@link Fact}.
+	 */
 	public static String getFactString(Fact fact) {
 		return getString(fact) + ".";
 	}
 
+	/**
+	 * Creates a String representation of a given {@link Constant}. Example: "c"
```
irina-dragoste: Explain what "c" means here.
A link to our Syntax page in the wiki would help.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -70,36 +86,92 @@ public static String getString(Literal literal) {
...
+	 * @param constant a {@link Constant}
+	 * @return String representation corresponding to a given {@link Constant}.
+	 */
 	public static String getString(Constant constant) {
 		return constant.getName();
 	}
 
+	/**
+	 * Creates a String representation of a given {@link ExistentialVariable}.
+	 * Example: "!X"
```
irina-dragoste: explain (for a variable name X, this method returns !X)

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -70,36 +86,92 @@ public static String getString(Literal literal) {
...
+	 * @return String representation corresponding to a given
+	 *         {@link ExistentialVariable}.
+	 */
 	public static String getString(ExistentialVariable existentialVariable) {
 		return "!" + existentialVariable.getName();
 	}
 
+	/**
+	 * Creates a String representation of a given {@link UniversalVariable}.
+	 * Example: "?X"
```
irina-dragoste: explain

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -70,36 +86,92 @@ public static String getString(Literal literal) {
...
+	 * 
+	 * @param universalVariable a {@link UniversalVariable}
+	 * @return String representation corresponding to a given
+	 *         {@link UniversalVariable}.
+	 */
 	public static String getString(UniversalVariable universalVariable) {
 		return "?" + universalVariable.getName();
 	}
 
+	/**
```
irina-dragoste: Explain

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -70,36 +86,92 @@ public static String getString(Literal literal) {
...
+	 * 
+	 * @param namedNull a {@link NamedNull}
+	 * @return String representation corresponding to a given {@link NamedNull}.
+	 */
 	public static String getString(NamedNull namedNull) {
 		return "_" + namedNull.getName();
 	}
 
+	/**
+	 * Creates a String representation of a given {@link Predicate}. Example: "p(2)"
```
irina-dragoste: explain

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -70,36 +86,92 @@ public static String getString(Literal literal) {
...
+	 * @return String representation corresponding to a given {@link Predicate}.
+	 */
 	public static String getString(Predicate predicate) {
-		return " Predicate [ name= " + predicate.getName() + ", arity= " + predicate.getArity() + "]";
+		return predicate.getName() + "(" + predicate.getArity() + ")";
 	}
 
+	/**
+	 * Creates a String representation of a given {@link DataSourceDeclaration}.
+	 * Example: "@source p(3): sparql(<https://example.org/sparql>, "var", "?var
```
irina-dragoste: explain

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -70,36 +86,92 @@ public static String getString(Literal literal) {
...
+	 */
 	public static String getString(DataSourceDeclaration dataSourceDeclaration) {
 		return "@source " + dataSourceDeclaration.getPredicate().getName() + "("
 				+ dataSourceDeclaration.getPredicate().getArity() + "): "
 				+ dataSourceDeclaration.getDataSource().getSyntacticRepresentation();
 	}
 
+	/**
+	 * Creates a String representation of a given {@link Conjunction}. Example:
+	 * "p(?X,?Y), ~q(a,?Z)"
```
irina-dragoste: explain

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -114,12 +186,28 @@ public static String getString(Conjunction<? extends Literal> conjunction) {
 		return stringBuilder.toString();
 	}
 
-	public static String getLanguageStringConstantName(LanguageStringConstant languageStringConstant) {
+	/**
+	 * Creates a String representation corresponding to the name of a given
+	 * {@link LanguageStringConstant}. Example: ""Test"@en"
```
irina-dragoste: explain

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -114,12 +186,28 @@ public static String getString(Conjunction<? extends Literal> conjunction) {
...
 		return "\"" + languageStringConstant.getString().replace("\\", "\\\\").replace("\"", "\\\"") + "\"@"
 				+ languageStringConstant.getLanguageTag();
 	}
 
-	public static String getDatatypeConstantName(DatatypeConstant datatypeConstant) {
+	/**
+	 * Creates a String representation corresponding to the name of a given
+	 * {@link DatatypeConstant}. Example: ""c"^^<http://example.org/mystring>"
+	 * 
+	 * @param datatypeConstant a {@link DatatypeConstant}
```
irina-dragoste: explain

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/EntityTest.java
irina-dragoste: You are right. Then maybe have the fields as local variables, if they are not used in other methods.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/CsvFileDataSource.java
```diff
@@ -73,4 +73,9 @@ public String toString() {
 		return "CsvFileDataSource [csvFile=" + getFile() + "]";
 	}
 
+	@Override
```
irina-dragoste: add javadoc

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/InMemoryDataSource.java
```diff
@@ -101,4 +101,10 @@ public String toConfigString() {
 		return null;
 	}
```
irina-dragoste: add javadoc

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/RdfFileDataSource.java
```diff
@@ -71,4 +71,9 @@ public String toString() {
 		return "RdfFileDataSource [rdfFile=" + getFile() + "]";
 	}
 
+	@Override
```
irina-dragoste: add Javadoc

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/SparqlQueryResultDataSource.java
```diff
@@ -164,4 +163,9 @@ public String toString() {
 				+ ", queryBody=" + queryBody + "]";
 	}
 
+	@Override
```
irina-dragoste: add javadoc

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/InMemoryDataSource.java
irina-dragoste: Should return the serialized content of this DataModel, which starts with a comment "serialising", and have one fact per line.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -129,6 +130,15 @@ public boolean equals(Object obj) {
 			return predicate.equals(other.predicate);
 		}
 
+		public String getSyntacticRepresentation() {
+			return Serializer.getString(this);
+		}
+
+		@Override
+		public String toString() {
```
irina-dragoste: don't override toString(), leave the one from Object be

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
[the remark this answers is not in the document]
irina-dragoste: throw MethodNotSupportedException

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/InMemoryDataSource.java
```diff
@@ -101,4 +101,10 @@ public String toConfigString() {
...
 
+	@Override
+	public String getSyntacticRepresentation() {
+		// TODO Auto-generated method stub
+		return null;
+	}
+
```
irina-dragoste: don't override toString()

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -201,8 +221,26 @@ public static String getConstantName(LanguageStringConstant languageStringConsta
...
+	 * @return String representation corresponding to a given
+	 *         {@link DatatypeConstant}.
+	 */
+	public static String getShortConstantName(DatatypeConstant datatypeConstant) {
+		if (datatypeConstant.getDatatype().equals(PrefixDeclarations.XSD_STRING)) {
+			return "\"" + datatypeConstant.getLexicalValue() + "\"";
+		} else {
+			return datatypeConstant.getLexicalValue();
+		}
```
mmarx: This should also handle:
  - LanguageStrings, which should serialise to `"string"@en`,
  - `xsd:integer`, which should serialise to `23`,
  - `xsd:decimal`, which should serialise to `1234.5678`, and
  - `xsd:double`, which should serialise to something like `12.345E67`.
Note that any of the numbers might be negative, and the exponent in doubles might have a different sign than the mantissa.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/SparqlQueryResultDataSource.java
alloka: we don't add javadoc for overriden methods right ? just once in the interface

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -49,8 +49,8 @@
 	public static final String existentialIdentifier = "!";
 	public static final String universalIdentifier = "?";
 	public static final String namedNullIdentifier = "_";
-	public static final String openBracket = "(";
-	public static final String closeBracket = ")";
+	public static final String openParentheses = "(";
+	public static final String closeParentheses = ")";
```
mmarx: `Parentheses` is the plural form, please use the singular `Parenthesis` here.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,260 @@
...
+	 * Creates a String representation corresponding to the name of a given
+	 * {@link LanguageStringConstant}.
+	 * 
+	 * @see <"https://github.com/knowsys/vlog4j/wiki"> for wiki.
+	 * @param languageStringConstant a {@link LanguageStringConstant}
+	 * @return String representation corresponding to the name of a given
+	 *         {@link LanguageStringConstant}.
+	 */
+	public static String getConstantName(LanguageStringConstant languageStringConstant) {
+		return "\"" + languageStringConstant.getString().replace("\\", "\\\\").replace("\"", "\\\"") + "\"@"
```
irina-dragoste: We could use have a method (for example, `private String addQuotations(String string)`) for `return "\""+sting+"\""`;

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,260 @@
...
+				+ languageStringConstant.getLanguageTag();
```
irina-dragoste: We could use a method (for example, `private String escape(String string)`)
for `return string.replace("\\", "\\\\").replace("\"", "\\\"")`

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,260 @@
...
+	 * {@link DatatypeConstant} without an IRI.
+	 * 
+	 * @see <"https://github.com/knowsys/vlog4j/wiki"> for wiki.
+	 * @param datatypeConstant a {@link DatatypeConstant}
+	 * @return String representation corresponding to a given
+	 *         {@link DatatypeConstant}.
+	 */
+	public static String getShortConstantName(DatatypeConstant datatypeConstant) {
+		if (datatypeConstant.getDatatype().equals(PrefixDeclarations.XSD_STRING)) {
+			return "\"" + datatypeConstant.getLexicalValue() + "\"";
```
irina-dragoste: Are we sure we don't need to escape `datatypeConstant.getLexicalValue(`) ?

alloka: I guess if we did, the string will look different and we don't want to do that, I'm keeping the other method as well in case it was needed

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/SparqlQueryResultDataSource.java
irina-dragoste: It depends, if you want to be specific make explicit what you change when you override, you should add a javadoc. I guess in this case it is not necessary, since we don't even have a specific wiki page for the syntax of each DataSource.

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/EntityTest.java
```diff
@@ -0,0 +1,190 @@
...
+		Variable x = Expressions.makeUniversalVariable("X");
+		Variable y = Expressions.makeUniversalVariable("Y");
+		Variable z = Expressions.makeExistentialVariable("Z");
+		NegativeLiteral atom1 = Expressions.makeNegativeLiteral("p", x, c);
+		PositiveLiteral atom2 = Expressions.makePositiveLiteral("p", x, y);
+		PositiveLiteral headAtom1 = Expressions.makePositiveLiteral("q", x, z);
+		Conjunction<Literal> bodyLiterals = Expressions.makeConjunction(atom1, atom2);
+		Conjunction<PositiveLiteral> headPositiveLiterals = Expressions.makePositiveConjunction(headAtom1);
+		Rule rule1 = new RuleImpl(headPositiveLiterals, bodyLiterals);
+		assertEquals(rule1, RuleParser.parseRule(rule1.toString()));
```
irina-dragoste: I would test 
assertEquals(bodyLiterals, RuleParser.parseRule(rule1.toString()).getBody()).
(and the same for the head conjunction).
Too bad that we don't have a package protected parseConjunction method to test.

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/EntityTest.java
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	@Test
+	public void literalToStringRoundTripTest() throws ParsingException {
+		Constant c = Expressions.makeAbstractConstant("c");
```
irina-dragoste: I would have two methods, one for testing positive literals, and one for testing negative literals.

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/EntityTest.java
```diff
@@ -0,0 +1,190 @@
...
+		Variable x = Expressions.makeUniversalVariable("X");
+		Variable z = Expressions.makeExistentialVariable("Z");
+		NegativeLiteral atom1 = Expressions.makeNegativeLiteral("p", x, c);
+		PositiveLiteral headAtom1 = Expressions.makePositiveLiteral("q", x, z);
+		Rule rule1 = Expressions.makeRule(headAtom1, atom1);
```
irina-dragoste: Why do we create a Rule, and not just use `RuleParser.parseLiteral` method?
I assume parseLiteral is called inside parseRule in the jj code.

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/EntityTest.java
```diff
@@ -0,0 +1,190 @@
...
+		assertEquals(rule1, RuleParser.parseRule(rule1.toString()));
+	}
+
+	@Test
+	public void datatypeDoubleConstantToStringRoundTripTest() throws ParsingException {
+		String shortDoubleConstant = "12.345E67";
+		assertEquals(shortDoubleConstant,
+				RuleParser.parseFact("p(\"12.345E67\"^^<http://www.w3.org/2001/XMLSchema#double>).").getArguments()
```
irina-dragoste: just reuse the variable `shortDoubleConstant` inside the other strings, otherwise we have to read each string and make sure it is identical to `shortDoubleConstant`

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/EntityTest.java
```diff
@@ -0,0 +1,190 @@
...
+						.get(0).toString());
+		assertEquals(shortDoubleConstant, RuleParser.parseFact("p(12.345E67).").getArguments().get(0).toString());
+	}
+
+	@Test
+	public void datatypeFloatConstantToStringRoundTripTest() throws ParsingException {
+		String floatConstant = "\"0.5\"^^<http://www.w3.org/2001/XMLSchema#float>";
+		assertEquals(floatConstant, RuleParser.parseFact("p(\"0.5\"^^<http://www.w3.org/2001/XMLSchema#float>).")
+				.getArguments().get(0).toString());
```
irina-dragoste: reuse the constant

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/EntityTest.java
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	@Test
+	public void datatypeStringConstantToStringRoundTripTest() throws ParsingException {
+		String shortStringConstant = "\"data\"";
+		assertEquals(shortStringConstant, RuleParser
+				.parseFact("p(\"data\"^^<http://www.w3.org/2001/XMLSchema#string>).").getArguments().get(0).toString());
```
irina-dragoste: reuse the constant

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/EntityTest.java
```diff
@@ -0,0 +1,190 @@
...
+		assertEquals(shortStringConstant, RuleParser.parseFact("p(\"data\").").getArguments().get(0).toString());
+	}
+
+	@Test
+	public void datatypeIntegerConstantToStringRoundTripTest() throws ParsingException {
+		String shortIntegerConstant = "1";
```
irina-dragoste: reuse constant

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/EntityTest.java
```diff
@@ -0,0 +1,190 @@
...
+		assertEquals(shortIntegerConstant, RuleParser.parseFact("p(\"1\"^^<http://www.w3.org/2001/XMLSchema#integer>).")
+				.getArguments().get(0).toString());
+		assertEquals(shortIntegerConstant, RuleParser.parseFact("p(1).").getArguments().get(0).toString());
+	}
+
+	@Test
+	public void datatypeDecimalToStringRoundTripTest() throws ParsingException {
```
irina-dragoste: reuse constant

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/EntityTest.java
```diff
@@ -0,0 +1,190 @@
...
+				"var", "?var wdt:P31 wd:Q5 .");
+		DataSourceDeclaration dataSourceDeclaration1 = new DataSourceDeclarationImpl(predicate1, dataSource);
+		RuleParser.parseInto(kb, dataSourceDeclaration1.toString());
+		assertEquals(dataSourceDeclaration1, kb.getDataSourceDeclarations().get(0));
+	}
+
+	@Test
+	public void rdfDataSourceDeclarationToStringParsingTest() throws ParsingException, IOException {
+		KnowledgeBase kb = new KnowledgeBase();
+		String INPUT_FOLDER = "src/test/data/input/";
```
irina-dragoste: this is a local variable, it should not have a constant (class public static final fields) naming convention.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -43,15 +43,22 @@
...
+	public static final String OPEN_PARENTHESIS = "(";
+	public static final String CLOSING_PARENTHESIS = ")";
+	public static final String RULE_SEPARATOR = " :- ";
+	public static final String AT = "@";
+	public static final String SOURCE = "@source ";
+	public static final String COLON = ": ";
+	public static final String CARET = "^";
+	public static final String LESS_THAN = "<";
+	public static final String MORE_THAN = ">";
+	public static final String ESCAPED_QUOTE = "\"";
```
mmarx: Why is this called `ESCAPED_QUOTE`? The String just contains a single quote character (the escaping happens on the Java syntax level here).

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/EntityTest.java
```diff
@@ -0,0 +1,190 @@
...
+		File unzippedRdfFile = new File(INPUT_FOLDER + "file.nt");
```
irina-dragoste: why not use only one variable to create a File?
(ex: `File unzippedRdfFile = new File("src/test/data/input/file.nt")`)

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/EntityTest.java
```diff
@@ -0,0 +1,190 @@
...
+		DataSourceDeclaration dataSourceDeclaration = new DataSourceDeclarationImpl(predicate1,
+				unzippedRdfFileDataSource);
+		RuleParser.parseInto(kb, dataSourceDeclaration.toString());
+		assertEquals(dataSourceDeclaration, kb.getDataSourceDeclarations().get(0));
+	}
+
+	@Test
+	public void csvDataSourceDeclarationToStringParsingTest() throws ParsingException, IOException {
+		KnowledgeBase kb = new KnowledgeBase();
+		String INPUT_FOLDER = "src/test/data/input/";
```
irina-dragoste: same as above

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/RuleParserTest.java
```diff
@@ -411,4 +412,24 @@ public void testBlankPredicateName() throws ParsingException {
 		RuleParser.parse(input);
 	}
 
+	@Test
+	public void DatatypeConstantgRoundTripTest() throws ParsingException {
```
irina-dragoste: This test does not do roundtrip.

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/RuleParserTest.java
```diff
@@ -411,4 +412,24 @@ public void testBlankPredicateName() throws ParsingException {
...
+		DatatypeConstantImpl datatypeConstantString = new DatatypeConstantImpl("data", PrefixDeclarations.XSD_STRING);
```
irina-dragoste: I would have a test method for each.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,271 @@
...
+	 * @see <"https://github.com/knowsys/vlog4j/wiki"> for wiki.
+	 * @param constant a {@link Constant}
+	 * @return String representation corresponding to a given {@link Constant}.
+	 */
+	public static String getString(Constant constant) {
+		return constant.getName();
+	}
+
+	public static String getString(DatatypeConstant constant) {
+		return getShortConstantName(constant);
```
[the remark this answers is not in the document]
alloka: The problem is that the original getConstantName is being used in some other places in code, so changing it will trigger alot of errors, so what do you see ?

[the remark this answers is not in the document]
alloka: for example, the reasoner and rdf

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/RuleParserTest.java
alloka: what do you suggest?, should we just remove all of them and the ones in the entitytest are enough ?

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Expressions.java
```diff
@@ -46,7 +46,19 @@
  * @author Markus Krötzsch
  *
  */
+
 public final class Expressions {
+
+	public static final String COLON_UNSPACED = ":";
```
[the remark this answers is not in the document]
alloka: sorry, this was a trial for something else and forgot to remove it.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	private static String escape(String string) {
+		return string.replace("\\", "\\\\").replace("\"", "\\\"");
+	}
+
+	private static String checkRelativeAbsoluteIri(String string) {
+		if ((string.contains(COLON_UNSPACED) || string.matches(INTEGER) || string.matches(DOUBLE)
+				|| string.matches(DECIMAL) || string.equals(TRUE) || string.equals(FALSE))
+				&& (!string.contains(LESS_THAN))) {
```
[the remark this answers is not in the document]
alloka: because we want to make sure that this doesn't have < > already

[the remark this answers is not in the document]
irina-dragoste: but it only checks for "<". So if you have "bla>", the toString will be "<bla>>".
Anyway, a "parsed" string that represents an IRI with never create an AbstractConstant or Predicate with name containing "<" or ">".  The parser does not decode "%3C", which is the URL encoding of "<", into "<".
In the future, if the IRI contains special predicates (like "<", for example), in the toString we would have to encode it (for example, replace it with "%3C"). This can be done in another task.

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/RuleParserTest.java
```diff
@@ -411,4 +412,20 @@ public void testBlankPredicateName() throws ParsingException {
 		RuleParser.parse(input);
 	}
 
+	@Test
+	public void predicateIriRoundTripTest() throws ParsingException {
+		AbstractConstantImpl a = new AbstractConstantImpl("a");
```
irina-dragoste: this is not a roundtrip test.
What do you want to test here?

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/RuleParserTest.java
```diff
@@ -411,4 +412,20 @@ public void testBlankPredicateName() throws ParsingException {
...
+		Fact f = RuleParser.parseFact("<1.e1>(a).");
+		Fact f2 = Expressions.makeFact("1.e1", a);
+		assertEquals(f, f2);
+	}
+
+	@Test
+	public void predicateIriRoundTripTest2() throws ParsingException {
+		AbstractConstantImpl a = new AbstractConstantImpl("a");
```
irina-dragoste: same as above

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/RuleParserTest.java
alloka: wanted to make sure its the same if i already have < > or not

alloka: but I guess this is already checked i will remove it

irina-dragoste: please check if this is indeed already tested, and if not, rename the unit test and leave it.

alloka: done

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -809,4 +809,4 @@ void setReasonerState(ReasonerState reasonerState) {
 		this.reasonerState = reasonerState;
 	}
 
-}
+}
```
irina-dragoste: There is no new line at the end of the file VLogReasoner. Please add the newline back.

Issue #20, opened 2018-04-16
OWL: support nominals


mkroetzsch:
Support for owl:oneOf is currently not implemented yet and this is not correctly reported with an exception. This should be fixed by supporting all admissible nominals in subclasses and superclasses. The architecture already has most of the functionality required, but variable substitutions should be added (rather than just relying on auxiliary facts as possible in subclasses.

irina-dragoste:
Axiom `SubClassOf(<A> ObjectOneOf(<a>))`
is incorrectly transformed into the rule` http://www.w3.org/2002/07/owl#Nothing(?X) :- A(?X)`

irina-dragoste:
Axioms `SubClassOf(ObjectOneOf(<a> <b>) <A>)` and `SubClassOf(ObjectOneOf(<a>) <A>)` are incorrectly transformed into the program 
```
rules: [A(?X) :- http://www.w3.org/2002/07/owl#Thing(?X)]
facts: []
```

irina-dragoste:
Partially fixed by #132: we support nominals in axioms of type subClassOf , with a ObjectOneOf class expression subtype. (ex: {a,b,c} subClassOf anyClassExpression). For any other nominal occurrences, we throw an exception.

The remaining fixes are in issue #142.

PR #132, opened 2019-11-11
Bug#20 nominals

irina-dragoste:
partial solution for bug #20 concerning adding support for nominals (i.e. ObjectOneOf class expressions): we only support nominals in axioms of type subClassOf , with a ObjectOneOf class expression subtype. (ex: `{a,b,c} subClassOf anyClassExpression`)


vlog4j-owlapi/src/main/java/org/semanticweb/vlog4j/owlapi/OwlAxiomToRulesConverter.java
```diff
@@ -219,14 +221,19 @@ void startAxiomConversion() {
...
-		this.startAxiomConversion();
-
-		final ClassToRuleHeadConverter headConverter = new ClassToRuleHeadConverter(this.frontierVariable, this);
-		superClass.accept(headConverter);
-		final ClassToRuleBodyConverter bodyConverter = new ClassToRuleBodyConverter(this.frontierVariable,
-				headConverter.body, headConverter.head, this);
-		bodyConverter.handleDisjunction(subClass, this.frontierVariable);
-		this.addRule(bodyConverter);
+		if (subClass instanceof OWLObjectOneOf) {
+			final OWLObjectOneOf subClassInstaceOf = (OWLObjectOneOf) subClass;
```
mmarx: `subClassInstanceOf` is misspelled.

vlog4j-owlapi/src/main/java/org/semanticweb/vlog4j/owlapi/OwlAxiomToRulesConverter.java
```diff
@@ -219,14 +221,19 @@ void startAxiomConversion() {
...
 	void addSubClassAxiom(final OWLClassExpression subClass, final OWLClassExpression superClass) {
-		this.startAxiomConversion();
-
-		final ClassToRuleHeadConverter headConverter = new ClassToRuleHeadConverter(this.frontierVariable, this);
-		superClass.accept(headConverter);
-		final ClassToRuleBodyConverter bodyConverter = new ClassToRuleBodyConverter(this.frontierVariable,
-				headConverter.body, headConverter.head, this);
-		bodyConverter.handleDisjunction(subClass, this.frontierVariable);
-		this.addRule(bodyConverter);
+		if (subClass instanceof OWLObjectOneOf) {
```
mmarx: I don't like the `instanceof` too much, but I guess it's quite hard to avoid here.

vlog4j-owlapi/src/main/java/org/semanticweb/vlog4j/owlapi/OwlAxiomToRulesConverter.java
irina-dragoste: I actually renamed it to `subClassObjectOneOf`, as it is more fitting.

PR #136, opened 2019-11-25
Prefer jacoco over cobertura

### LINKED ISSUE #85: cobertura does not support Java 1.8
Test coverage reports are limited due to a number of failures that can be seen in the TravisCI logs. Many are caused by the missing support for Java 1.8 constructs such as `default` in the cobertura plugin.

The problem could be solved by switching to JaCoCo for code coverage, which, however, is not an insignificant effort in terms of configuration.

mmarx:
Replace cobertura by jacoco.

Fixes #85.


mmarx:
Okay, excludes are not yet working correctly.

Issue #135, opened 2019-11-21
Core: obtain materialised chase


mkroetzsch:
There should be a way of getting access to all inferred facts, without having to issue queries for each predicate. This could be presented as a stream, collection, or iterator over Literal objects (to be discussed), and it could be provided as a single method of the Reasoner, or possibly through a dependent object ("Chase") that the Reasoner can return and through which this (and other) modes of access are possible.

irina-dragoste:
Possible solutions:

- [ ] export all facts into a file in our RLS syntax, encoded as facts
- [ ] export all facts into a set of CSVDataSourceDeclarations for each IDB and/or EDB.

irina-dragoste:
note that vlog4j-base class karmaresearch.vlog.VLog already supports method

```
/**
     * Creates a CSV file at the specified location, for the specified
     * predicate.
     *
     * @param predicateName
     *            the predicate name
     * @param fileName
     *            the location
     * @exception NotStartedException
     *                is thrown when vlog is not started yet, or materialization
     *                has not run yet
     * @exception IOException
     *                is thrown when the file could not be written for some
     *                reason
     */
    public native void writePredicateToCsv(String predicateName,
            String fileName) throws NotStartedException, IOException;

```

irina-dragoste:
Solution:
- Write all chase facts into an OutputStream in VLog4j Fact syntax
- export all facts chase (without duplicates), not just inferred facts
- avoid creating Fact objects from query results, but have a String representation of query results that is the same as for the corresponding Facts. 
- unit tests could check that the exported chase result is readable
- for handling the String representation of a QueryResult and its Predicate, use Serializer class / a different class than VLogReasoner
- vLog- specific dependencies should not be in a Serializer class

PR #149, opened 2020-01-08
Obtain inferred facts

alloka:
#135


vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +811,34 @@ ReasonerState getReasonerState() {
...
+				}
+			}
+		}
+		stream.close();
+
+	}
+
+	@Override
+	public void getKbFacts(String filePath) throws IOException {
+		// TODO Auto-generated method stub
```
irina-dragoste: remove TODO

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +811,34 @@ ReasonerState getReasonerState() {
...
+			for (PositiveLiteral positiveLiteral : rule.getHead()) {
+				headLiterals.add(positiveLiteral);
+			}
+		}
+		for (PositiveLiteral positiveliteral : headLiterals) {
+			try (final QueryResultIterator queryAnswers = this.answerQuery(positiveliteral, true)) {
+				while (queryAnswers.hasNext()) {
+					QueryResult queryAnswer = queryAnswers.next();
+					stream.write((positiveliteral.getPredicate().getName()
+							+ queryAnswer.getTerms().toString().replace("[", "(").replace("]", ").") + "\n")
```
irina-dragoste: We should not have QueryResult String processing in Reasoner implementation.
Instead, this should be done in  QueryResult#toString() (which could delegate to Serializer).
Also, instead of replacing "[" with "(" in the String representation of a List, we should construct the String ourselves while iterating trough the list elements, and use the "(" and ")" constants from Serializer class.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +811,34 @@ ReasonerState getReasonerState() {
 	void setReasonerState(ReasonerState reasonerState) {
 		this.reasonerState = reasonerState;
 	}
+
+	@Override
+	public void getKbFacts(OutputStream stream) throws IOException {
```
irina-dragoste: Method behaviour is inconsistent with its javadoc: the facts explicitly added in the KB may not appear in the rule heads, therefore method writes only the inferred facts.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +811,34 @@ ReasonerState getReasonerState() {
...
+	@Override
+	public void getKbFacts(OutputStream stream) throws IOException {
+		// TODO Auto-generated method stub
+		HashSet<PositiveLiteral> headLiterals = new HashSet<PositiveLiteral>();
+		for (Rule rule : this.knowledgeBase.getRules()) {
+			for (PositiveLiteral positiveLiteral : rule.getHead()) {
+				headLiterals.add(positiveLiteral);
+			}
+		}
+		for (PositiveLiteral positiveliteral : headLiterals) {
```
irina-dragoste: you may write the same fact several times: consider rules
```
a(x) ->b(x,x)
b(x,y)->b(y,x)
```
Or even
```
a(x) -> b(x)
a(y) -> b(y)
```
Where b(x) and b(y) are different PositiveLiterals.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +811,34 @@ ReasonerState getReasonerState() {
...
+			try (final QueryResultIterator queryAnswers = this.answerQuery(positiveliteral, true)) {
+				while (queryAnswers.hasNext()) {
```
irina-dragoste: you can use java8 syntax here for a shorter code

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +811,34 @@ ReasonerState getReasonerState() {
...
+		}
+		stream.close();
+
+	}
+
+	@Override
+	public void getKbFacts(String filePath) throws IOException {
+		// TODO Auto-generated method stub
+		OutputStream stream = new FileOutputStream(filePath);
+		getKbFacts(stream);
```
irina-dragoste: same comments about incorrect behaviour as above

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/Reasoner.java
```diff
@@ -86,6 +87,22 @@ static Reasoner getInstance() {
 	 */
 	KnowledgeBase getKnowledgeBase();
 
+	/**
+	 * Exports all the facts of the knowledge base to an OutputStream. This includes
+	 * the added facts and the inferred facts as well.
+	 * 
+	 * @param an OutpumStream for the facts to be written to.
+	 */
+	void getKbFacts(OutputStream stream) throws IOException;
```
irina-dragoste: usually methods named "get" return something, maybe rename to "writeKBFacts", or even better, "writeReasoningResult"?

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -384,9 +386,13 @@ private static String addQuotes(final String string) {
 		return QUOTE + string + QUOTE;
 	}
 
-
 	private static String addAngleBrackets(final String string) {
 		return LESS_THAN + string + MORE_THAN;
 	}
 
+	public static String getFactOutput(Predicate predicate, List<Term> terms) {
+		return predicate.getName() + terms.toString().replace(OPENING_BRACKET, OPENING_PARENTHESIS)
```
irina-dragoste: Also, when serializing a Fact, we do more changes (call `getIRIString` on the predicate name).
Please make sure the result of the chase can be round-tripped (i.e. read as a set of equivalent facts).
This may require more processing when the result includes blanks.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +812,42 @@ ReasonerState getReasonerState() {
 	void setReasonerState(ReasonerState reasonerState) {
 		this.reasonerState = reasonerState;
 	}
+
+	@Override
+	public void writeInferredFacts(OutputStream stream) throws IOException {
+		HashSet<Predicate> toBeQueriedHeadPredicates = new HashSet<Predicate>();
```
irina-dragoste: It is a best practice to use interfaces for variable references, i.e
```
Set<Predicate> toBeQueriedHeadPredicates = new HashSet<Predicate>();
List<Term> tobeGroundedVariables = new ArrayList<Term>();
```

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +812,42 @@ ReasonerState getReasonerState() {
...
+		for (Fact fact : this.knowledgeBase.getFacts()) {
+			if (!toBeQueriedHeadPredicates.contains(fact.getPredicate()))
+				stream.write((fact.toString() + "\n").getBytes());
+		}
+		for (Predicate predicate : toBeQueriedHeadPredicates) {
+			ArrayList<Term> tobeGroundedVariables = new ArrayList<Term>();
+			for (int i = 0; i < predicate.getArity(); i++) {
+				tobeGroundedVariables.add(Expressions.makeUniversalVariable("X" + i));
+			}
+			final QueryResultIterator answers = this
```
irina-dragoste: use try-with-resources for the `QueryResultIterator `
(for example, look for how is use it in the vlog4j-examples module)

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +812,42 @@ ReasonerState getReasonerState() {
...
+					.answerQuery(Expressions.makePositiveLiteral(predicate, tobeGroundedVariables), true);
+			answers.forEachRemaining(queryAnswer -> {
+				try {
+					stream.write(Serializer.getFactOutput(predicate, queryAnswer.getTerms()).getBytes());
+				}
+
+				catch (IOException e) {
```
irina-dragoste: never do `e.printStackTrace();` when catching an exception, it is too expensive.
We should propagate this exception (like we do for exportQueryAnswersToCsv), and perhaps log it first.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +812,42 @@ ReasonerState getReasonerState() {
...
+					e.printStackTrace();
+				}
+			});
+		}
+		stream.close();
+	}
+
+	@Override
+	public void writeInferredFacts(String filePath) throws IOException {
+		OutputStream stream = new FileOutputStream(filePath);
```
irina-dragoste: as OutputStream is auto-closeable, use try-with-resources.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +812,42 @@ ReasonerState getReasonerState() {
...
+
+	@Override
+	public void writeInferredFacts(OutputStream stream) throws IOException {
+		HashSet<Predicate> toBeQueriedHeadPredicates = new HashSet<Predicate>();
+		for (Rule rule : this.knowledgeBase.getRules()) {
+			for (Literal literal : rule.getHead()) {
+				toBeQueriedHeadPredicates.add(literal.getPredicate());
+			}
+		}
+		for (Fact fact : this.knowledgeBase.getFacts()) {
```
irina-dragoste: what about the "facts" that come from a DataSource?
They will not appear in the chase result.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/Reasoner.java
```diff
@@ -86,6 +87,20 @@ static Reasoner getInstance() {
 	 */
 	KnowledgeBase getKnowledgeBase();
 
+	/**
+	 * Exports the inferred facts of the knowledge base to an OutputStream.
+	 * 
+	 * @param an OutputStream for the facts to be written to.
+	 */
+	void writeInferredFacts(OutputStream stream) throws IOException;
```
irina-dragoste: I thought we want to obtain the whole chase result, not only the inferred facts.
Anyway, the method returns all inferred facts plus some facts that were added to the KB as Fact objects.
If we only want to obtain inferred facts, we should not return explicit facts; if we want to obtain the entire chase result, we should add all explicit facts (i.e. the ones added to the KB from a `DataSource`).

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/RuleParserTest.java
```diff
@@ -438,4 +454,82 @@ public void testCustomDatatype() throws ParsingException {
...
+		final DataSource dataSource3 = new SparqlQueryResultDataSource(new URL("https://example.org/"), "var2",
+				"?var2 wdt:P31 wd:Q5 .");
+		final DataSourceDeclaration dataSourceDeclaration3 = new DataSourceDeclarationImpl(predicate2, dataSource3);
+		final Predicate predicate4 = Expressions.makePredicate("city", 1);
+		final DataSourceDeclaration dataSourceDeclaration4 = new DataSourceDeclarationImpl(predicate4, dataSource2);
+		RuleParser.parseInto(kb, fact.toString() + ".");
+		RuleParser.parseInto(kb, dataSourceDeclaration1.toString());
+		RuleParser.parseInto(kb, dataSourceDeclaration2.toString());
+		RuleParser.parseInto(kb, dataSourceDeclaration3.toString());
+		RuleParser.parseInto(kb, dataSourceDeclaration4.toString());
```
irina-dragoste: you can just add these Statements (fact, dataSourceDeclaration1...4) to the kb, with addStatement, you don't need to call toString on them, and then parse them into equivalent objects, and them add those equiv statements to the kb.

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/RuleParserTest.java
```diff
@@ -438,4 +454,82 @@ public void testCustomDatatype() throws ParsingException {
...
+				+ "country(germany). \n" //
+				+ "university(tudresden, germany). \n" //
+				+ "streetAddress(tudresden, \"Mommsenstraße 9\", \"01069\", \"Dresden\") . \n" //
+				+ "zipLocation(\"01069\", dresden) . \n" //
+				+ "locatedIn(?X,?Y) :- location(?X,?Y) . \n" //
+				+ "locatedIn(?X,?Z) :- locatedIn(?X,?Y), locatedIn(?Y,?Z) . \n" //
+				+ "address(?Uni, ?Street, ?ZIP, ?City) :- streetAddress(?Uni, ?Street, ?ZIP, ?CityName), zipLocation(?ZIP, ?City) . \n"
+				+ "address(?Uni, !Street, !ZIP, !City), locatedIn(!City, ?Country) :- university(?Uni, ?Country) . \n";
+		RuleParser.parseInto(kb, rules);
+		final String facts = "location(germany,europe). \n" //
```
irina-dragoste: the beginning of this string is identical to a part of a previous string.
just reuse a variable in this case, it is confusing to check that every single character is the same in two different strings.

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/RuleParserTest.java
```diff
@@ -438,4 +454,82 @@ public void testCustomDatatype() throws ParsingException {
 		DatatypeConstant result = (DatatypeConstant) literal.getConstants().toArray()[0];
 		assertEquals(constant, result);
 	}
+
+	@Test
+	public void testGetFacts() throws ParsingException, IOException {
+		KnowledgeBase kb = new KnowledgeBase();
+		final DataSource dataSource1 = new SparqlQueryResultDataSource(new URL("https://example.org/"), "var",
```
irina-dragoste: we should not use SparqlQueryResultDataSources in unit tests, as the tests may fail if the internet connection is low.

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/RuleParserTest.java
```diff
@@ -438,4 +454,82 @@ public void testCustomDatatype() throws ParsingException {
 		DatatypeConstant result = (DatatypeConstant) literal.getConstants().toArray()[0];
 		assertEquals(constant, result);
 	}
+
+	@Test
+	public void testGetFacts() throws ParsingException, IOException {
```
irina-dragoste: instead of hardcoding the expected chase result, for an integration test I suggest the following scenario:
1. you add some data+rules in a KB `kb1`
2. you reason on `kb1`
3. you query `kb1` for every single predicate you have in the kb, and save all of these queries&results in a collection `res` (for example, a HashMap)
4. you export the chase result of  `kb1` to a file, and then parse it into  `kb2` 
5. you reason on `kb2` (note that there are no rules in `kb2`)
6. you query `kb2`  for all the queries in `res`, and check that you get the same results as when you queried `kb1`.
Blanks introduced to satisfy existential rules may require some pre-processing, I am not sure if they will be round-tripped into AbstractConstants.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/Reasoner.java
alloka: @irina-dragoste  I query over all predicates, actually isn't the facts obtained somehow from datasources are either part of the facts of the knowledge base which I obtain or at least they have a predicate which I query for ?, are u sure this is not covered ?

alloka: @irina-dragoste  nvm done

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/RuleParserTest.java
alloka: maybe we can discuss this in the meeting because I get the feeling that this is the same as implementing the method all over again, and I don't see why what we have now is not ok

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -384,9 +386,15 @@ private static String addQuotes(final String string) {
 		return QUOTE + string + QUOTE;
 	}
 
-
 	private static String addAngleBrackets(final String string) {
 		return LESS_THAN + string + MORE_THAN;
 	}
 
+	public static String getFactOutput(Predicate predicate, List<Term> terms) {
```
irina-dragoste: I don't know if the word "output" makes sense in the method name. Why not call it getString or getFactString ?

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/model/implementation/Serializer.java
```diff
@@ -384,9 +386,15 @@ private static String addQuotes(final String string) {
...
+		return getIRIString(predicate.getName())
```
irina-dragoste: the existing method `public static String getString(final Literal literal) {` can delegate to this method (once implemented correctly), so code can be reused.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/Reasoner.java
```diff
@@ -86,6 +87,20 @@ static Reasoner getInstance() {
 	 */
 	KnowledgeBase getKnowledgeBase();
 
+	/**
+	 * Exports all the facts of the knowledge base to an OutputStream.
+	 * 
+	 * @param an OutputStream for the facts to be written to.
+	 */
+	void writeFacts(OutputStream stream) throws IOException;
```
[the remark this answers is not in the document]
irina-dragoste: "writeInferences"

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +812,47 @@ ReasonerState getReasonerState() {
...
+			for (int i = 0; i < predicate.getArity(); i++) {
+				tobeGroundedVariables.add(Expressions.makeUniversalVariable("X" + i));
+			}
+			try (final QueryResultIterator answers = this
+					.answerQuery(Expressions.makePositiveLiteral(predicate, tobeGroundedVariables), true)) {
+				answers.forEachRemaining(queryAnswer -> {
+					try {
+						stream.write(Serializer.getFactOutput(predicate, queryAnswer.getTerms()).getBytes());
+					} catch (IOException e) {
+						throw new RuntimeException();
```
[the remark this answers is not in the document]
alloka: sorry forgot to put the e inside the runtime exception

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/RuleParserTest.java
```diff
@@ -438,4 +448,37 @@ public void testCustomDatatype() throws ParsingException {
 		DatatypeConstant result = (DatatypeConstant) literal.getConstants().toArray()[0];
 		assertEquals(constant, result);
 	}
+
+	@Test
+	public void testWriteInferences() throws ParsingException, IOException {
+		KnowledgeBase kb = new KnowledgeBase();
+		final InMemoryDataSource locations = new InMemoryDataSource(2, 1);
+		locations.addTuple("dresden", "germany");
+		RuleParser.parseInto(kb, fact.toString() + ".");
```
irina-dragoste: why does `fact` need to be parsed, and cannot be directly added to the KB?

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +812,46 @@ ReasonerState getReasonerState() {
...
+		}
+		for (DataSourceDeclaration dataSourceDeclaration : this.knowledgeBase.getDataSourceDeclarations()) {
+			toBeQueriedHeadPredicates.add(dataSourceDeclaration.getPredicate());
+		}
+		for (Fact fact : this.knowledgeBase.getFacts()) {
+			toBeQueriedHeadPredicates.add(fact.getPredicate());
+		}
+
+		for (Predicate predicate : toBeQueriedHeadPredicates) {
+			List<Term> tobeGroundedVariables = new ArrayList<Term>();
```
irina-dragoste: inconsistent naming "toBe" versus "tobe"; first version is better, since it is camelCase.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +812,46 @@ ReasonerState getReasonerState() {
...
+			for (int i = 0; i < predicate.getArity(); i++) {
+				tobeGroundedVariables.add(Expressions.makeUniversalVariable("X" + i));
+			}
+			try (final QueryResultIterator answers = this
+					.answerQuery(Expressions.makePositiveLiteral(predicate, tobeGroundedVariables), true)) {
+				answers.forEachRemaining(queryAnswer -> {
+					try {
+						stream.write(Serializer.getFactString(predicate, queryAnswer.getTerms()).getBytes());
+					} catch (IOException e) {
+						throw new RuntimeException(e);
```
irina-dragoste: is this exception turned into an unchecked exception because it is in a forEachRemaning function? Then, it is better to use a non-java8 way of iterating, so that the checked exception can be directly propagated.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +812,46 @@ ReasonerState getReasonerState() {
...
+			toBeQueriedHeadPredicates.add(fact.getPredicate());
+		}
+
+		for (Predicate predicate : toBeQueriedHeadPredicates) {
+			List<Term> tobeGroundedVariables = new ArrayList<Term>();
+			for (int i = 0; i < predicate.getArity(); i++) {
+				tobeGroundedVariables.add(Expressions.makeUniversalVariable("X" + i));
+			}
+			try (final QueryResultIterator answers = this
+					.answerQuery(Expressions.makePositiveLiteral(predicate, tobeGroundedVariables), true)) {
```
irina-dragoste: We should not call answerQuery directly, as it creates too many objects (for each `TermQueryResultIterator`, it creates a `VLogQueryResultIterator` to include the `Correctness` result, but for  `writeInferedFacts`, the `Correctness` result is the same for all querries. 
Also, logWarningOnCorrectness() is called for every querried predicate. If `Correctness `is not `SOUND_AND_COMPLETE` and WARNING logging is enabled, a warning will be logged for each queried predicate, which is both slow and annoying.

Also, `answerQuery` does checks for each query that only need to be done once (reasoner not closed, kb loaded) or never (query not null)  for  `writeInferedFacts` .

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/Reasoner.java
```diff
@@ -86,6 +87,20 @@ static Reasoner getInstance() {
 	 */
 	KnowledgeBase getKnowledgeBase();
 
+	/**
+	 * Exports all the facts of the knowledge base to an OutputStream.
+	 * 
+	 * @param an OutputStream for the facts to be written to.
+	 */
+	void writeInferences(OutputStream stream) throws IOException;
```
irina-dragoste: method shoud return `Correctness `(see `answerQuery`, it is the same logic).

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/Reasoner.java
```diff
@@ -86,6 +87,24 @@ static Reasoner getInstance() {
 	 */
 	KnowledgeBase getKnowledgeBase();
 
+	/**
+	 * @return the correctness of the query answers, depending on the state of the
+	 *         reasoning (materialisation) and its {@link KnowledgeBase} Exports all
+	 *         the (explicit and implicit) facts of the knowledge base to an
```
irina-dragoste: What the method does is written in the return statement, instead of above

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/Reasoner.java
```diff
@@ -86,6 +87,24 @@ static Reasoner getInstance() {
...
+	 */
+	Correctness writeInferences(OutputStream stream) throws IOException;
+
+	/**
+	 * Exports all the (explicit and implicit) facts of the knowledge base to a
+	 * desired file.
+	 * 
+	 * @param a String of the file path for the facts to be written to.
+	 */
+	void writeInferences(String filePath) throws IOException;
```
irina-dragoste: should also return Correctness (and update its javadoc in consequence)

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +813,59 @@ ReasonerState getReasonerState() {
...
+		for (Predicate predicate : toBeQueriedHeadPredicates) {
+			List<Term> toBeGroundedVariables = new ArrayList<Term>();
+			for (int i = 0; i < predicate.getArity(); i++) {
+				toBeGroundedVariables.add(Expressions.makeUniversalVariable("X" + i));
+			}
+			final karmaresearch.vlog.Atom vLogAtom = ModelToVLogConverter
+					.toVLogAtom(Expressions.makePositiveLiteral(predicate, toBeGroundedVariables));
+			try {
+				stringQueryResultIterator = this.vLog.query(vLogAtom, true, false);
+				answers = new VLogQueryResultIterator(stringQueryResultIterator, this.correctness);
```
irina-dragoste: it is not necessary to do this. Why is a new VLogQueryResultIterator created? one can use the initial vlog iterator, and call `VLogToModelConverter#toTermList` on the term list that `next()` returns.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +813,59 @@ ReasonerState getReasonerState() {
 	void setReasonerState(ReasonerState reasonerState) {
 		this.reasonerState = reasonerState;
 	}
+
+	@Override
+	public Correctness writeInferences(OutputStream stream) throws IOException {
+		QueryResult queryAnswer;
```
irina-dragoste: why declare this variable here and change its value, instead as locally in the try block? It is not necessary and also makes code more unreadable.
Also, variables should be final unless it is absolutely necessary to change their values. 
Reasoner#answerQuery() declares variable `stringQueryResultIterator ` that is only initialised in a try block, as this variable is later returned. Here,  you don't need the value of `queryAnswer ` outside the try block.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +813,59 @@ ReasonerState getReasonerState() {
 	void setReasonerState(ReasonerState reasonerState) {
 		this.reasonerState = reasonerState;
 	}
+
+	@Override
+	public Correctness writeInferences(OutputStream stream) throws IOException {
+		QueryResult queryAnswer;
+		Set<Predicate> toBeQueriedHeadPredicates = new HashSet<Predicate>();
+		TermQueryResultIterator stringQueryResultIterator;
```
irina-dragoste: same as above with `queryAnswer`

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +813,59 @@ ReasonerState getReasonerState() {
 	void setReasonerState(ReasonerState reasonerState) {
 		this.reasonerState = reasonerState;
 	}
+
+	@Override
+	public Correctness writeInferences(OutputStream stream) throws IOException {
+		QueryResult queryAnswer;
+		Set<Predicate> toBeQueriedHeadPredicates = new HashSet<Predicate>();
+		TermQueryResultIterator stringQueryResultIterator;
+		QueryResultIterator answers;
```
irina-dragoste: same as above with `queryAnswer`.

Generally, declare a variable just before you use it, this makes code more readable.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +813,59 @@ ReasonerState getReasonerState() {
...
+			for (Literal literal : rule.getHead()) {
+				toBeQueriedHeadPredicates.add(literal.getPredicate());
+			}
+		}
+		for (DataSourceDeclaration dataSourceDeclaration : this.knowledgeBase.getDataSourceDeclarations()) {
+			toBeQueriedHeadPredicates.add(dataSourceDeclaration.getPredicate());
+		}
+		for (Fact fact : this.knowledgeBase.getFacts()) {
+			toBeQueriedHeadPredicates.add(fact.getPredicate());
+		}
```
irina-dragoste: it woulc be more readable if this was separated in a private/package-protected method.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +813,59 @@ ReasonerState getReasonerState() {
...
+
+		for (Predicate predicate : toBeQueriedHeadPredicates) {
+			List<Term> toBeGroundedVariables = new ArrayList<Term>();
+			for (int i = 0; i < predicate.getArity(); i++) {
+				toBeGroundedVariables.add(Expressions.makeUniversalVariable("X" + i));
+			}
+			final karmaresearch.vlog.Atom vLogAtom = ModelToVLogConverter
+					.toVLogAtom(Expressions.makePositiveLiteral(predicate, toBeGroundedVariables));
```
irina-dragoste: it woulc be more readable if this was separated in a private/package-protected method that returns a vlog4j PositiveLiteral.

ModelToVLogConverter.toVLogAtom can be called on that returned PositiveLiteral.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +813,59 @@ ReasonerState getReasonerState() {
...
+			try {
```
irina-dragoste: why not use try-with-resources?

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
irina-dragoste: this comment was not addressed.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
irina-dragoste: This method should perform the same validations on the reasoner state as `answerQuery `and `exportQueryAnswersToCsv`

vlog4j-parser/src/test/java/org/semanticweb/vlog4j/syntax/parser/RuleParserTest.java
```diff
@@ -438,4 +448,37 @@ public void testCustomDatatype() throws ParsingException {
 		DatatypeConstant result = (DatatypeConstant) literal.getConstants().toArray()[0];
 		assertEquals(constant, result);
 	}
+
+	@Test
+	public void testWriteInferences() throws ParsingException, IOException {
```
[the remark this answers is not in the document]
irina-dragoste: issue was not addressed

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +813,59 @@ ReasonerState getReasonerState() {
...
+				}
+			} catch (final NotStartedException e) {
+				throw new RuntimeException("Inconsistent reasoner state.", e);
+			} catch (final NonExistingPredicateException e1) {
+				LOGGER.warn("Query uses predicate " + predicate
+						+ " that does not occur in the knowledge base. Answer must be empty!");
+			}
+
+		}
+		return this.correctness;
```
irina-dragoste: this method should behave similarily as answerQuery and exportQueryAnswersToCsv, and log a warning for incorrect answers.

PR #151, opened 2020-02-07
Count query results natively

larryjgonzalez:
The native API of [VLog](https://github.com/karmaresearch/vlog) now supports the counting of query results. 
This pull request uses this new functionality. It is not necessary anymore to count query results in VLog4j (ExamplesUtils.iteratorSize), but call reasoner.queryAnswerSize instead.

In order to test this pull request, please build the latest VLog, using at least this [commit](6b62f413d8f3a27ef153fe2df549f3293faec61a)


vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -594,6 +594,32 @@ public QueryResultIterator answerQuery(PositiveLiteral query, boolean includeBla
...
+
+	@Override
+	public long queryAnswerSize(PositiveLiteral query, boolean includeNulls) {
+		validateNotClosed();
+		Validate.notNull(query, "Query atom must not be null!");
+
+		final boolean filterBlanks = !includeNulls;
+		final karmaresearch.vlog.Atom vLogAtom = ModelToVLogConverter.toVLogAtom(query);
+
+		long result = -1;
```
irina-dragoste: I would not initialize result here, and initialize it at line 618 with value 0, instead of `return 0`.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -808,4 +834,4 @@ ReasonerState getReasonerState() {
 	void setReasonerState(ReasonerState reasonerState) {
 		this.reasonerState = reasonerState;
 	}
-}
+}
```
irina-dragoste: please do not remove the newline character at the end of a .java file, it is required (for parsing, I think)

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/Reasoner.java
```diff
@@ -331,6 +331,67 @@ static Reasoner getInstance() {
...
+	 * <br>
+	 * An answer to the query is the terms a fact that matches the {@code query}:
+	 * the fact predicate is the same as the {@code query} predicate, the
+	 * {@link TermType#CONSTANT} terms of the {@code query} appear in the answer
+	 * fact at the same term position, and the {@link TermType#VARIABLE} terms of
+	 * the {@code query} are matched by terms in the fact, either named
+	 * ({@link TermType#CONSTANT}) or anonymous ({@link TermType#NAMED_NULL}). The
+	 * same variable name identifies the same term in the answer fact. <br>
+	 * 
+	 * Depending on the state of the reasoning (materialisation) and its
```
irina-dragoste: we discuss Correctness in the javadoc, but `queryAnswerSize()` method does not return in any way its correctness result. We should either return Correctness together with the answer size, or ignore it in the javadoc, or suggest that Correctness can be found out together with the query result in `answerQuery` or `exportQueryAnswers`

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -594,6 +594,32 @@ public QueryResultIterator answerQuery(PositiveLiteral query, boolean includeBla
...
 
+	@Override
+	public long queryAnswerSize(PositiveLiteral query) {
+		return queryAnswerSize(query, true);
+	}
+
+	@Override
+	public long queryAnswerSize(PositiveLiteral query, boolean includeNulls) {
+		validateNotClosed();
+		Validate.notNull(query, "Query atom must not be null!");
```
irina-dragoste: the method should behave like `answerQuery`, i.e. throw an exception if the KB has not been loaded.

vlog4j-examples/src/main/java/org/semanticweb/vlog4j/examples/CountingTriangles.java
```diff
@@ -61,15 +61,18 @@ public static void main(final String[] args) throws IOException {
...
 			// Due to symmetry, each triangle is found six times, hence we divide by 6:
 			System.out.println("The number of triangles of countries that mutually border each other was "
-					+ (ExamplesUtils.getQueryAnswerCount("triangle(?X,?Y,?Z)", reasoner) / 6) + ".");
+					+ (triangles / 6) + ".");
 		}
 
 	}
-}
+
+}
```
irina-dragoste: please do not remove this last empty line in a java file

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/reasoner/implementation/QueryAnswerSizeTest.java
```diff
@@ -0,0 +1,287 @@
...
+	private static final Fact factPd = Expressions.makeFact(predP, d);
+
+	private static final Fact factQe = Expressions.makeFact(predQ, e);
+	private static final Fact factQf = Expressions.makeFact(predQ, f);
+
+	private static final PositiveLiteral Rdy = Expressions.makePositiveLiteral(predR, d, y);
+	private static final PositiveLiteral Rey = Expressions.makePositiveLiteral(predR, e, y);
+	private static final PositiveLiteral Rxd = Expressions.makePositiveLiteral(predR, x, d);
+	private static final PositiveLiteral Rxe = Expressions.makePositiveLiteral(predR, x, e);
+
```
irina-dragoste: It should be tested that queries without variables return correct results.

For example,
 for a KB with fact ` f=P(c)`, `queryAnswerSize(f)` returns 1.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
mmarx: For what it's worth, it's “required” because POSIX mandates that every line needs a terminating LF, and some really old tools may not work correctly on files not having those (as far as I know, we don't use any of them; line counts using `wc` might be off by one, but that's about it).

As long as we're still happily mixing UNIX-style LF terminators with Windows-style CRLF terminators in the same codebase, I wouldn't put too much emphasis on sticking to such antiquated requirements.

irina-dragoste: Ok, I see.

larryjgonzalez:
The changes requested have been implemented. Please check the new implementation.

vlog4j-client/src/main/java/org/semanticweb/vlog4j/client/picocli/VLog4jClientMaterialize.java
```diff
@@ -230,8 +230,7 @@ private void doSaveQueryResults(final Reasoner reasoner, final PositiveLiteral q
 	}
 
 	private void doPrintResults(final Reasoner reasoner, final PositiveLiteral query) {
-		System.out.println(
-				"Number of query answers in " + query + ": " + ClientUtils.getQueryAnswerCount(query, reasoner));
+		System.out.println("Number of query answers in " + query + ": " + reasoner.queryAnswerSize(query));
```
irina-dragoste: now, this will print what `toString()` value of the returned `QueryAnswersSize` object. As `toString()` is not overwritten in `QueryAnswersSize` class, the default implementation from `Object` class will be called, giving an  undreadable hash code.
I suggest to print `reasoner.queryAnswerSize(query).getSize()` and `reasoner.queryAnswerSize(query).getCorrectness()`  here.

 `QueryAnswersSize` should also overwrite `toString `to provide information about `size `and also `correctness`.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/Reasoner.java
```diff
@@ -331,6 +331,66 @@ static Reasoner getInstance() {
...
+	 * @param query        a {@link PositiveLiteral} representing the query to be
+	 *                     answered.
+	 * @param includeNulls if {@code true}, {@link QueryResult}s containing terms of
+	 *                     type {@link TermType#NAMED_NULL} (representing anonymous
+	 *                     individuals introduced to satisfy rule existentially
+	 *                     quantified variables) will be included. Otherwise, the
+	 *                     answers will only contain the {@link QueryResult}s with
+	 *                     terms of type {@link TermType#CONSTANT} (representing
+	 *                     named individuals).
+	 * @return number of facts in the extension of the query.
```
irina-dragoste: update the return statement in the Javadoc, now the returned value contains both the number of facts and the Correctness.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/implementation/VLogReasoner.java
```diff
@@ -594,6 +595,35 @@ public QueryResultIterator answerQuery(PositiveLiteral query, boolean includeBla
...
+		try {
+			result = this.vLog.querySize(vLogAtom, true, filterBlanks);
+		} catch (NotStartedException e) {
+			throw new RuntimeException("Inconsistent reasoner state.", e);
+		} catch (NonExistingPredicateException e) {
+			LOGGER.warn("Query uses predicate " + query.getPredicate()
+					+ " that does not occur in the knowledge base. Answer must be empty!");
+			result = 0;
+		}
+		return new QueryAnswersSize(this.correctness, result);
```
irina-dragoste: like in `answerQuery` and `exportQueryAnswersToCsv`, a warning must be logged if Correctness is not `SOUND_AND_COMPLETE`.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/QueryAnswersSize.java
```diff
@@ -0,0 +1,64 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+
+/**
+ * 
+ * @author Larry González
+ *
+ */
```
irina-dragoste: the Javadoc is missing information about this class.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/QueryAnswersSize.java
```diff
@@ -0,0 +1,64 @@
...
+public class QueryAnswersSize {
+
+	final Correctness correctness;
+	final long size;
+
+	public QueryAnswersSize(Correctness correctness, int size) {
```
irina-dragoste: this constructor is not necessary, as `long` variables can be assigned `int` values.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/QueryAnswersSize.java
```diff
@@ -0,0 +1,64 @@
...
+		this.correctness = correctness;
+		this.size = size;
+	}
+
+	public QueryAnswersSize(Correctness correctness, long size) {
```
irina-dragoste: missing javaodc

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/QueryAnswersSize.java
```diff
@@ -0,0 +1,64 @@
...
+	 * results may be unsound.
+	 * </ul>
+	 * 
+	 * @return query result correctness
+	 */
+	public Correctness getCorrectness() {
+		return this.correctness;
+	}
+
+	public long getSize() {
```
irina-dragoste: missing javadoc

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/reasoner/implementation/QueryAnswerSizeTest.java
```diff
@@ -0,0 +1,293 @@
...
+			assertEquals(2, reasoner.queryAnswerSize(Px, true).getSize());
+			assertEquals(2, reasoner.queryAnswerSize(Px, false).getSize());
+			assertEquals(2, reasoner.queryAnswerSize(Qx, true).getSize());
+			assertEquals(2, reasoner.queryAnswerSize(Qx, false).getSize());
+			assertEquals(0, reasoner.queryAnswerSize(Rxy, true).getSize());
+			assertEquals(0, reasoner.queryAnswerSize(Rxy, false).getSize());
+			assertEquals(1, reasoner.queryAnswerSize(factPc, false).getSize());
+			assertEquals(1, reasoner.queryAnswerSize(factPd, false).getSize());
+			assertEquals(1, reasoner.queryAnswerSize(factQc, false).getSize());
+			assertEquals(1, reasoner.queryAnswerSize(factQd, false).getSize());
```
irina-dragoste: I would also add checks for `queryAnswerSize(fact,true)`

vlog4j-core/src/test/java/org/semanticweb/vlog4j/core/reasoner/implementation/QueryAnswerSizeTest.java
```diff
@@ -0,0 +1,293 @@
...
+		}
+	}
+
```
irina-dragoste: I would also add tests for a query with no variables, for which
- the predicate is not in the KB
- the predicate is in the KB, but there is no query answer
(this is more of a test for vLog then a unit test of our method, but it doesn't hurt)

larryjgonzalez: The first three test cover these cases

irina-dragoste: but in these first three tests, the queries have variables. I was referring to queries that only contain constants.

vlog4j-core/src/main/java/org/semanticweb/vlog4j/core/reasoner/Reasoner.java
```diff
@@ -358,6 +359,65 @@ static Reasoner getInstance() {
 	 */
 	QueryResultIterator answerQuery(PositiveLiteral query, boolean includeNulls);
 
+	/*
```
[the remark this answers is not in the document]
irina-dragoste: also, the javadoc should contain a body, not just @param and @return statements

PR #163, opened 2020-03-02
Core: Make FileDataSource serialization OS-independent

### LINKED ISSUE #140: Core: DatasourceDeclaration toString is OS dependent for FileDataSources.

mmarx:
Fixes #140.

### DIFF
--- a/rulewerk-core/src/main/java/org/semanticweb/rulewerk/core/model/implementation/Serializer.java
+++ b/rulewerk-core/src/main/java/org/semanticweb/rulewerk/core/model/implementation/Serializer.java
@@ -380,7 +380,7 @@ public final class Serializer {
 	}
 
 	private static String getFileString(final FileDataSource fileDataSource) {
-		return getString(fileDataSource.getFile().toString());
+		return getString(fileDataSource.getPath().toString());
 	}
 
 	private static String getIRIString(final String string) {
--- a/rulewerk-core/src/main/java/org/semanticweb/rulewerk/core/reasoner/implementation/FileDataSource.java
+++ b/rulewerk-core/src/main/java/org/semanticweb/rulewerk/core/reasoner/implementation/FileDataSource.java
@@ -9,9 +9,9 @@ package org.semanticweb.rulewerk.core.reasoner.implementation;
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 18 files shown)


rulewerk-core/src/main/java/org/semanticweb/rulewerk/core/reasoner/implementation/FileDataSource.java
```diff
@@ -54,32 +55,33 @@
...
 	 *                                  does not occur in
 	 *                                  {@code possibleExtensions}.
 	 */
-	public FileDataSource(@NonNull final File file, final Iterable<String> possibleExtensions) throws IOException {
-		Validate.notNull(file, "Data source file cannot be null!");
-		final String fileName = file.getName();
-
-		this.file = file;
+	public FileDataSource(final String filePath, final Iterable<String> possibleExtensions) throws IOException {
+		Validate.notBlank(filePath, "Data source file name cannot be null!");
```
irina-dragoste: validation message is not updated

rulewerk-core/src/main/java/org/semanticweb/rulewerk/core/model/implementation/Serializer.java
```diff
@@ -380,7 +380,7 @@ public static String getString(final SparqlQueryResultDataSource dataSource) {
 	}
 
 	private static String getFileString(final FileDataSource fileDataSource) {
-		return getString(fileDataSource.getFile().toString());
+		return getString(fileDataSource.getPath().toString());
```
irina-dragoste: getPath() already returns a String

rulewerk-core/src/main/java/org/semanticweb/rulewerk/core/reasoner/implementation/FileDataSource.java
```diff
@@ -54,32 +55,33 @@
...
+
+		this.file = new File(filePath);
+		this.filePath = filePath.replaceAll("\\\\", "/"); // canonicalise windows-style path separators
+		this.fileName = this.filePath.substring(this.filePath.lastIndexOf("/") + 1); // just the file name
```
irina-dragoste: as `\\` is a valid string occurring in a Unix file name, fileName and filePath would lead to undesired behaviour. 
fileName should be obtained from the File object, and filePath that is returned by getPath() and used in the serialization should be unmodified, as it is user input.

rulewerk-core/src/main/java/org/semanticweb/rulewerk/core/reasoner/implementation/FileDataSource.java
```diff
@@ -54,32 +55,33 @@
...
 	 */
-	public FileDataSource(@NonNull final File file, final Iterable<String> possibleExtensions) throws IOException {
-		Validate.notNull(file, "Data source file cannot be null!");
-		final String fileName = file.getName();
-
-		this.file = file;
+	public FileDataSource(final String filePath, final Iterable<String> possibleExtensions) throws IOException {
+		Validate.notBlank(filePath, "Data source file name cannot be null!");
+
+		this.file = new File(filePath);
```
irina-dragoste: Do we need to keep the File field after the constructor was called?

mmarx: It's still used in `equals()` and `hashCode()`, for Example.

irina-dragoste: I see, I guess we still want to consider two FileDataSource objects with different path names that refer to the same file as equal.

PR #179, opened 2020-06-17
Fast query result iterator

mkroetzsch:
Replace VLogQueryResultIterator by faster implementation
* New class avoids use of VLog's TermQueryResultIterator
* Hence no intermediate string representations of results
* Two forms of caching reduce dictionary lookups and number of term objects in RuleWerk


rulewerk-vlog/src/main/java/org/semanticweb/rulewerk/reasoner/vlog/VLogFastQueryResultIterator.java
```diff
@@ -0,0 +1,234 @@
...
+	 * of repeatedly used terms, since the cache entries there are always pushed out
+	 * before being needed again. This implementation will at least cache a maximal
+	 * initial fragment in such cases. It is also faster to write and requires less
+	 * memory.
+	 * 
+	 * @author Markus Kroetzsch
+	 *
+	 */
+	static class OrderedTermCache {
+		final private HashMap<Long, Term> terms = new HashMap<>();
```
mmarx, line 82: It might make sense to use `maxCapacity` as the initial capacity of the `HashMap`, to avoid repeatedly growing the map.

rulewerk-vlog/src/main/java/org/semanticweb/rulewerk/reasoner/vlog/VLogFastQueryResultIterator.java
```diff
@@ -0,0 +1,234 @@
...
+	final OrderedTermCache termCache;
+
+	private final Correctness correctness;
+
+	/**
+	 * Create a new {@link VLogFastQueryResultIterator}.
+	 * 
+	 * @param queryResultIterator
+	 * @param materialisationState
+	 * @param vLog
```
mmarx, line 147: Missing Javadoc

rulewerk-vlog/src/main/java/org/semanticweb/rulewerk/reasoner/vlog/VLogFastQueryResultIterator.java
```diff
@@ -0,0 +1,234 @@
...
+	@Override
+	public QueryResult next() {
+		final Term[] terms;
+		long[] idTuple = vLogQueryResultIterator.next();
+		terms = new Term[idTuple.length];
+
+		if (firstResult) {
+			resultSize = terms.length;
+			prevIds = new long[resultSize - 1];
+			prevTerms = new Term[resultSize - 1];
```
mmarx, line 172: Consider using `lastIndex = terms.length - 1` instead, which makes the following code much more readable.

rulewerk-vlog/src/main/java/org/semanticweb/rulewerk/reasoner/vlog/VLogFastQueryResultIterator.java
```diff
@@ -0,0 +1,234 @@
...
+			// TODO: the string operation to make null names should possibly be provided by
+			// VLog rather than being hardcoded here?
+			if (s == null) {
+				return new NamedNullImpl("" + (id >> 40) + "_" + ((id >> 32) & 0377) + "_" + (id & 0xffffffffL));
+			} else {
+				return VLogToModelConverter.toConstant(s);
+			}
+		} catch (NotStartedException e) {
+			// Should not happen, we just did a query ...
+			throw new RuntimeException(e);
```
mmarx, line 220: throw `RulewerkRuntimeException` instead.

rulewerk-vlog/src/main/java/org/semanticweb/rulewerk/reasoner/vlog/VLogReasoner.java
```diff
@@ -158,18 +158,18 @@ void load() throws IOException {
...
+			break;
+		case KB_LOADED:
+		case MATERIALISED:
+			// do nothing, all KB is already loaded
+			break;
+		case KB_CHANGED:
+			resetReasoner();
+			loadKnowledgeBase();
+		default:
+			break;
```
mmarx, line 172: Please drop this hunk.

rulewerk-vlog/src/main/java/org/semanticweb/rulewerk/reasoner/vlog/VLogReasoner.java
```diff
@@ -328,23 +328,23 @@ public boolean reason() throws IOException {
...
+		case KB_CHANGED:
+			resetReasoner();
+			load();
+			runChase();
+			break;
+		case MATERIALISED:
+			runChase();
+			break;
+		default:
+			break;
```
mmarx, line 347: Please drop this hunk.

rulewerk-vlog/src/main/java/org/semanticweb/rulewerk/reasoner/vlog/VLogReasoner.java
```diff
@@ -400,7 +403,40 @@ public QueryResultIterator answerQuery(PositiveLiteral query, boolean includeNul
...
+	}
+
+	/**
+	 * Utility method copied from {@link karmaresearch.vlog.VLog}.
+	 * 
+	 * @FIXME This should be provided by VLog and made visible to us rather than
+	 *        being copied here.
+	 * @param terms
+	 * @return
+	 * @throws NotStartedException
```
mmarx, line 416: Missing Javadoc

rulewerk-vlog/src/main/java/org/semanticweb/rulewerk/reasoner/vlog/VLogFastQueryResultIterator.java
mkroetzsch: Not sure if we should have such big maps by default. In the case I worked with, e.g., a cache size of around 78 was enough to have 100% hits upon second encounter of an id. That was for a total of around 1500 ids and around 7 million cache lookups.

rulewerk-vlog/src/main/java/org/semanticweb/rulewerk/reasoner/vlog/VLogFastQueryResultIterator.java
mkroetzsch: Can you explain? The last index of the prev* arrays is actually resultSize-2, so that name would seem confusing (but maybe I misread the comment).

rulewerk-vlog/src/main/java/org/semanticweb/rulewerk/reasoner/vlog/VLogFastQueryResultIterator.java
mkroetzsch: Why do we actually have RulewerkRuntimeException? Note that this line will never be executed in practice either.

rulewerk-vlog/src/main/java/org/semanticweb/rulewerk/reasoner/vlog/VLogFastQueryResultIterator.java
mmarx: We could also use something smaller, then. The default for `HashMap` is 16, though, which is probably too small.

rulewerk-vlog/src/main/java/org/semanticweb/rulewerk/reasoner/vlog/VLogFastQueryResultIterator.java
mmarx: Actually, it should rather throw `ReasonerStateException` (which `RulewerkRuntimeException` is a base class of); this is inconsistently used throughout the code base, however.

rulewerk-vlog/src/main/java/org/semanticweb/rulewerk/reasoner/vlog/VLogFastQueryResultIterator.java
mmarx: I was thinking about the last index of `terms`, but then `lastTerm` might be more specific and less confusing.

rulewerk-vlog/src/main/java/org/semanticweb/rulewerk/reasoner/vlog/VLogFastQueryResultIterator.java
mkroetzsch: It will need to grow, indeed, but I doubt that this has a measurable performance impact (the time taken by, say, 10 hashmap space duplications should be very small, and we would probably not initialise at a much larger size either).

rulewerk-vlog/src/main/java/org/semanticweb/rulewerk/reasoner/vlog/VLogFastQueryResultIterator.java
mkroetzsch: Using an index (ordinal) as a size (quantity) does not seem so clear to me. After sleeping over it, I don't think this code needs any rewriting to be of acceptable quality ;-).

mmarx: Feel free to use a more appropriate name, but the point I object to is that you're using `resultSize - 1` in multiple places, specifically in `i < resultSize - 1` comparisons that are essentially off-by-one errors waiting to happen, so that should be the value stored in a variable.

PR #196, opened 2020-12-07
Reliances

larry-gonzalez:
Functional prototype to compute Positive Reliances and Restraining Relations


coverage/pom.xml
```diff
@@ -14,46 +14,46 @@
...
+		<dependency>
+			<groupId>${project.groupId}</groupId>
+			<artifactId>rulewerk-commands</artifactId>
+			<version>${project.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>${project.groupId}</groupId>
+			<artifactId>rulewerk-client</artifactId>
+			<version>${project.version}</version>
+		</dependency>
```
mmarx, line 56: This needs a dependency on `rulewerk-reliances` as well, cf. the comment in the main `pom.xml`; otherwise the coverage aggregation will not work correctly.

larryjgonzalez: Thank you for the review. It is added now in a new commit

PR #204, opened 2021-03-29
Integration tests (old Issues)

larryjgonzalez:
(Description edited after discussion)

Branch to start integration-tests. 

I have created the maven module `org.semanticweb.rulewerk.integrationtests` that contains a single *test* package called `vlogissues`, that enclose some classes that reproduce an issue reported in [VLog](https://github.com/karmaresearch/vlog/) in a black-box way.

Please note:
* To avoid delays in development, some tests have been marked with `@Ignore` flag


mkroetzsch:
Tests that are meant to test VLog should be placed in the new module for integration tests (as discussed elsewhere). Tests in the vlog-rulewerk module should only test Java code that is part of Rulewerk (ideally, the vlog-rulewerk module should not have test failures even if VLog C++ is not working at all).

Issue #122, opened 2019-10-11
OWL: OWLDisjointObjectPropertiesAxiom support


irina-dragoste:
OWLDisjointObjectPropertiesAxioms are currently not supported by the owl to rule conveter.

PR #212, opened 2021-08-28
#122 owlapi owl disjoint object properties axiom support

irina-dragoste:
handles #122.


rulewerk-owlapi/src/main/java/org/semanticweb/rulewerk/owlapi/OwlToRulesConversionHelper.java
```diff
@@ -97,19 +98,23 @@ public static Predicate getConjunctionAuxiliaryClassPredicate(final Collection<O
...
 
-	private static String getAuxiliaryClassName(final String prefix,
-			final Collection<OWLClassExpression> owlClassExpressions) {
+	private static String getAuxiliaryOWLObjectName(final String prefix,
+			final Collection<? extends OWLObject> owlObjects) {
 		final MessageDigest messageDigest;
 		try {
 			messageDigest = MessageDigest.getInstance("MD5");
-			for (final OWLClassExpression owlClassExpression : owlClassExpressions) {
+			for (final OWLObject owlClassExpression : owlObjects) {
```
mmarx, line 117: Since this is no longer necessarily a class expression, the variable name should be changed as well.

mmarx (changes requested):
Apart from the no-longer adequately named variable, this looks good to me.

rulewerk-owlapi/src/main/java/org/semanticweb/rulewerk/owlapi/OwlToRulesConversionHelper.java
irina-dragoste: done.

PR #225, opened 2022-12-14
Feature/mvn2nix

mmarx:
Use `mvn2nix` instead of the fragile double-building approach for obtaining the rulewerk dependencies repository. This properly locks all versions of the dependencies, making the build reproducible, and the lock file can automatically be updated when the `pom.xml` files change (the output should be sorted, however, to avoid spurious conflicts).

Also split out the VLog integration tests into a separate invocation of `maven-failsafe-plugin`, so that a failure in those does not force the build to fail.


mmarx:
Another todo: check if failing VLog integration tests still fail the VLog workflow.

irina-dragoste (approved):

